PR #37, opened 2014-09-02
Ignore javajet copied to org.talend.designer.codegen/resources during de...

RyanSkraba:
...velopment.

These files are copied from other repositories to this location when running within eclipse, and show up when doing a "git status".  This is an exceptional location, and these files should never be modified or checked in here.


nrousseau:
not needed anymore

PR #38, opened 2014-09-03
[TDI-30511] Add FIRE_TRIGGER option to tMSSqlBulkExec

brazabr:
This change allow the user to flag the BULK INSERT operation to execute triggers in tMSSqlBulkExec component.


pteyssier:
This is integrated in https://github.com/Talend/tdi-studio-se/pull/1856

PR #42, opened 2014-09-26
tOracleInput and tPostgresqlInput fail with NPE on null Geometry

fxprunayre:
See https://github.com/talend-spatial/talend-spatial/issues/60


fxprunayre:
Closing it has been applied - see JIRA.

PR #58, opened 2014-12-02
TMDM-7936: only the "Fire a Create / Update event" option is ticked will...

sbliu:
this modification should also be merged to branch 6.0


glzhou:
Fixed in TDI-32145

PR #69, opened 2015-03-03
TDQ-10039 copy tMatchGroup work well.

qiongli:


ggu-talend:
Yes, before there are some problem for external nodes.

PR #76, opened 2015-03-12
TDQ-7943: because we Remove tLaunchDQReport from the palette, so remove

msjian:
its related migration task


ggu-talend:
Shouldn't delete the migration task simply like this.

PR #135, opened 2015-09-21
TBD-2564 - Migration task for split local/cluster mode in Spark Configuration

ftom:


rdubois:
Closed because changes were required. Will be pushed directly.

PR #160, opened 2015-11-19
TIPAAS-6691:tActionFailure component

csuntalendbj:
Please merge the changes for the tActionFailure Component ,so that the tLogCatcher will support tActionFailure.


atelesh:
>> SHOW_IF="IS_STUDIO_EE_VERSION == 'true'"
we need to find another solution since:
PO>> We need to use the same condition when we show/not show tAction* components.

PR #161, opened 2015-11-19
TDI-34289: Bugfixes and data type settings for EXASol components

jlolling:
1. tEXABulkExec:Allowing to use various number formats.
Add to the settings a schema based widget to setup the formats.
Changed the API and use different methods for Date, Number and other
types.
2. tEXABulkExec: Bug fixed: without external connection no commit
3. Change the MVN url of the jdbc driver in all EXASol components


jlolling:
I will close this pull request because it contains useless a commit.

PR #175, opened 2015-12-21
TMDM-9111 MDM transaction handling is wrong in multithreaded environment

htyin:
Generate transaction for each thread.


glzhou:
As agreed earlier, It's better to use ThreadLocal to share transaction within a thread.

PR #179, opened 2016-01-13
TESB-16985:tESBConsumer using default SOAPAction header when there is no SOAPAction defined

csuntalendbj:
pls merge for this issue .


main/plugins/org.talend.designer.maven.job/resources/templates/standalone/pom_job_template.xml
```diff
@@ -53,6 +53,14 @@
 
 	<build>
 		<finalName>${talend.job.finalName}</finalName>
+		<resources>
+            <resource>
+                <directory>${sourcecodes.dir}</directory>
```
ggu-talend: Because this property come from one profile, if not active, will be problem

PR #198, opened 2016-03-01
TDI-35280 bug: Talend stores te content of a bean to the wrong file!

cmeng-talend:
https://jira.talendforge.org/browse/TDI-35280


ycbai:
It's OK.

PR #217, opened 2016-03-10
TUP-4129:TCOMP:unwanted property appears on the component properties view(TUP-4010)

hcyi:


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java
```diff
@@ -536,6 +536,9 @@ public void addMainParameters(final List<ElementParameter> listParam, INode node
         param.setReadOnly(true);
         listParam.add(param);
 
+        // TUP-4129
+        ComponentProperties compProperties = ComponentsUtils.getComponentProperties(getName());
+        ComponentService service = ComponentsUtils.getComponentService();
         param = new ElementParameter(node);
```
francisu: The ComponentProperties should not be obtained newly; they should be the properties already associated with the node. Also, there is no reason to pass the service into this method.
I'm going to fix these issues.

PR #227, opened 2016-03-14
TDI-35382: tWaitForfile with trigger action "file created" is not

mswei-talend:
triggered if a filename is used twice
https://jira.talendforge.org/browse/TDI-35382
Note: Add some code to update the cache in every loop.


pteyssier:
You can rewrite it as equivalent to an intersection:
- avoid the negation for a bit more clarity
- assign/replace the temporary intersection list to the original list

PR #244, opened 2016-03-16
TUP-4145:Set initial value of an ENUM to be first possible value(TUP-4010)

hcyi:


hcyi:
With a wrong ticket number.

PR #248, opened 2016-03-16
TDI-35391 : The option 'sap jco 2.*' from tSAPConnection whence it's

pxmiao-talend:
done
https://jira.talendforge.org/browse/TDI-35391

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tSAPCommit/tSAPCommit_main.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tSAPCommit/tSAPCommit_main.javajet
@@ -19,10 +19,10 @@ imports="
 
     String conn = "conn_" + connection;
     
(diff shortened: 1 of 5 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tSAPCommit/tSAPCommit_main.javajet
```diff
@@ -19,10 +19,10 @@ imports="
 
     String conn = "conn_" + connection;
     
-    String VERSION_SAP2 = "sapjco.jar";
+    
     String VERSION_SAP3 = "sapjco3.jar";
     
-    String version = VERSION_SAP2;
+    String version = null;
```
pteyssier: I don't think the version should be set to null as there was a default value before. Minor but could you set it to VERSION_SAP3 please?

PR #252, opened 2016-03-18
TDI-33821 TDI-34895: update tGroovy and tGroovyFile components to groovy version 2.4.6

stavytskyi:


main/plugins/org.talend.libraries.groovy/.classpath
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
jzhao-talend: @stavytskyi 
1)"org.talend.libraries.groovy" should not be removed.
It is used in tBonitaDeploy/tBonitaInstantiateProcess.

 2)The groovy jar is APL2. So I think it is better included in the studio.

PR #254, opened 2016-03-18
TUP-4244:Missing the module name when drag&drop component(TUP-4010)

hcyi:


ycbai:
Hi please use SchemaUtils.getMetadataTables() instead of  for (ModelElement modelElement : connection.getOwnedElement()) ...

PR #266, opened 2016-03-23
TUP-4152:ComponentsProvider is scary(TUP-4010)

hcyi:


ycbai:
I think it is not a good way to only create a abstract class and use its static methods.

PR #267, opened 2016-03-24
Bugfix/6.2/tdi 35602

pxmiao-talend:


pteyssier:
This pull request concerns also a commit on oracle bulk component. Furthermore this commit was the correct ones on oracle.

PR #297, opened 2016-04-01
TESB-17183:Implement the function of export route with spring boot in studio

csuntalendbj:
this is a new feature of spring boot .pls merge


nrousseau:
Where this microservice type will be displayed?  From the code modified, seems it will be used for any kind of job, is that right?
And seems even if there is no any route.

main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/DefaultRunProcessService.java
```diff
@@ -185,6 +188,19 @@ protected IProcessor createJavaProcessor(IProcess process, Property property, bo
...
+            Bundle bundle = Platform.getBundle(PluginChecker.EXPORT_ROUTE_PLUGIN_ID);
+            try {
+                Class camelJavaProcessor = bundle.loadClass("org.talend.resources.export.maven.runprocess.CamelJavaProcessor");
+
+                Constructor constructor = camelJavaProcessor.getConstructor(IProcess.class, Property.class, boolean.class);
+
+                return (MavenJavaProcessor) constructor.newInstance(process, property, filenameFromLabel);
+
+            } catch (Exception e) {
+                e.printStackTrace();
```
nrousseau: no catch of error? usually use use at least ExceptionHandler.process(e), at least to catch the error and have it in the log.

csuntalendbj: sorry ,I will add this.

csuntalendbj:
microservice will display when users want to export the route as microservie (a jar file with spring dependency).

PR #307, opened 2016-04-08
TUP-4296:String type Property's default value missing double quote when job close and reopen(TUP-4010)

hcyi:


dlin-talend:
cann't reproduce .

PR #326, opened 2016-04-13
TUP-4216: Build job without Items checked produces broken artifact.

zwxue:


test/plugins/org.talend.repository.test/META-INF/MANIFEST.MF
```diff
@@ -4,4 +4,5 @@ Bundle-Name: Test
 Bundle-SymbolicName: org.talend.repository.test
 Bundle-Version: 6.2.0.qualifier
 Fragment-Host: org.talend.repository
-Require-Bundle: org.talend.testutils
+Require-Bundle: org.talend.testutils,
+ org.talend.designer.core
```
nrousseau: test plugin should never require any other bundle normally.
This cause a big problem of cycle dependency, since eclipse will consider org.talend.designer.core use org.talend.repository, and org.talend.repository (through test plugin) uses org.talend.designer.core

PR #341, opened 2016-04-19
TCOMP-128: Fixed runtime of tSalesforceGetUpdated and tSalesforceGetDeleted 

jzhao-talend:
Parse String value to Date for Date Property


jzhao-talend:
Have pushed.

PR #344, opened 2016-04-21
TDI-35834: Salesforce connection failed when job migrate to new framework

jzhao-talend:
https://jira.talendforge.org/browse/TDI-35834
1)Fixed password migration problem job and metadata
2)Module migration problem
3)Use connection migration problem


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java
```diff
@@ -989,7 +989,11 @@ public String getCodegenValue(Property property, String value) {
             return "\"" + value + "\"";//$NON-NLS-1$ //$NON-NLS-2$ 
         }
         if (property.getType() == Property.Type.ENUM) {
-            return "\"" + value + "\"";//$NON-NLS-1$ //$NON-NLS-2$
+        	if(value.indexOf("context.") > -1 || value.indexOf("globalMap.get") > -1){
```
jzhao-talend: This changes is for customer module support context and globaleMap value in old framework.
I am not sure there is a better way to handle this.

main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/controller/GenericHiddenTextController.java
```diff
@@ -59,6 +60,12 @@ protected boolean isReadOnly() {
     @Override
     public Control createControl(Composite subComposite, IElementParameter param, int numInRow, int nbInRow, int top,
             Control lastControl) {
+    	if(param.isRepositoryValueUsed() || isInWizard()){
+    		String pwd = CryptoHelper.getDefault().decrypt(String.valueOf(param.getValue()));
```
jzhao-talend: This is for metadata migration, when I set the real password value to the password field.
When I reopen the wizard connection the password was encrypt. I don't know why.
And not sure there is other way to fixed this issue. Need confirm with studio side.

PR #408, opened 2016-05-12
Bugfix/master/tdi 35569

nrousseau:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/TalendGridLayer.java
```diff
@@ -32,6 +33,8 @@
 
     public static final Color GRID_COLOR = ColorConstants.black;
 
+    public static boolean reSet = true;
+
```
nrousseau: not used? remove if it's not even used after please

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/connections/ConnectionFigure.java
```diff
@@ -425,5 +427,25 @@ public void setLineWidth(int w) {
...
+                int expand = (int) (getLineWidthFloat() / 2.0f);
+                bounds = getPoints().getBounds().getExpanded(expand, expand);
+            for (int i = 0; i < getChildren().size(); i++) {
+                IFigure child = (IFigure) getChildren().get(i);
+                bounds.union(child.getBounds());
+            }
+        }
+        return bounds;
+//        return super.getBounds();
+    }
```
nrousseau: no use to keep some comments, even more in a new code

PR #421, opened 2016-05-21
TDI-36277 change log level for technical info

jlolling:


jlolling:
These commits can also be applied to the release 6.1.1 without any changes. That is what I need for my customer requirement.

PR #423, opened 2016-05-23
Feature/6.2.1/tbd 3475 new parameters4 hive wizard

cmeng-talend:
https://jira.talendforge.org/browse/TBD-3475

depend on https://github.com/Talend/tcommon-studio-se/pull/445


ycbai:
It's OK.

PR #431, opened 2016-05-25
TDI-35892 : globalMap.get("tMysqlOutput_1_ERROR_MESSAGE") return null

pxmiao-talend:
without 'die on error' on the component
https://jira.talendforge.org/browse/TDI-35892


main/plugins/org.talend.designer.components.localprovider/components/tMysqlOutput/tMysqlOutput_end.javajet
```diff
@@ -400,6 +400,7 @@ skeleton="../templates/db_output_bulk.skeleton"
 	    	    	}
 					dbLog.logPrintedException("e.getMessage()");
 					%>
+					globalMap.put(currentComponent+"_ERROR_MESSAGE",e.getMessage());
```
pteyssier: There is only one place to fix?

PR #438, opened 2016-05-27
TMDM-9111 : MDM transaction handling is wrong in multithreaded enviro…

lpw82gf:
…nment


pbabin:
There might be performance improvements regarding the way to get the sessionId from backend.
While asking creation of a new transaction from server (in MDMTransactionClient.newTransaction method), we could retrieve that sessionId from the response. Thus it would reduce the Get call only when "useClientTranId" parameter is used. Also com.talend.mdm.transaction.client.MDMTransaction object can ultimately hold the value in both cases so commit() or rollback() methods do not need an input parameter, and sessionID does not need to be part of the global Map object.

PR #496, opened 2016-06-19
TDI-36630 tFTPClose

jlolling:
This component should be maintained by Talend because it is actually an mandatory functionality and the implementation highly depending on the implementation of tFTPConnection.


pteyssier:
Thanks for your contribution Jan, we fixed this in https://jira.talendforge.org/browse/TDI-31839

PR #501, opened 2016-06-20
TDI-36651: fix the runtime part for the return property

wwang-talend:
https://jira.talendforge.org/browse/TDI-36651

### DIFF
--- a/main/plugins/org.talend.designer.codegen/jet_stub/generic/component_end.javajet
+++ b/main/plugins/org.talend.designer.codegen/jet_stub/generic/component_end.javajet
@@ -49,20 +49,20 @@ if(resultMap_<%=cid%>!=null) {
 	for(java.util.Map.Entry<String,Object> entry_<%=cid%> : resultMap_<%=cid%>.entrySet()) {
 		switch(entry_<%=cid%>.getKey()) {//for the old key
 		case org.talend.components.api.component.ComponentDefinition.RETURN_ERROR_MESSAGE :
-			globalMap.put("<%=cid%>_ERROR_MESSAGE", entry_<%=cid%>.getValue());
+			container_<%=cid%>.setComponentData("<%=cid%>", "ERROR_MESSAGE", entry_<%=cid%>.getValue());
 			break;
 		case org.talend.components.api.component.ComponentDefinition.RETURN_TOTAL_RECORD_COUNT :
-			globalMap.put("<%=cid%>_NB_LINE", entry_<%=cid%>.getValue());
+			container_<%=cid%>.setComponentData("<%=cid%>", "NB_LINE", entry_<%=cid%>.getValue());
 			break;
 		case org.talend.components.api.component.ComponentDefinition.RETURN_SUCCESS_RECORD_COUNT :
-			globalMap.put("<%=cid%>_NB_SUCCESS", entry_<%=cid%>.getValue());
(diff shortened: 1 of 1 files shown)


wwang-talend:
merge it and we can adjust it later as we need to make it work now for test.

main/plugins/org.talend.designer.codegen/jet_stub/generic/component_end.javajet
```diff
@@ -33,13 +33,36 @@ if (metadata == null) {
...
+%>
+    org.talend.components.api.component.runtime.Result resultObject_<%=cid%> = (org.talend.components.api.component.runtime.Result)writer_<%=cid%>.close();
+    final java.util.Map<String, Object> resultMap_<%=cid%> = writer_<%=cid%>.getWriteOperation().finalize(java.util.Arrays.<org.talend.components.api.component.runtime.Result>asList(resultObject_<%=cid%>), container_<%=cid%>);
+<%
 }
 %>
+if(resultMap_<%=cid%>!=null) {
+	for(java.util.Map.Entry<String,Object> entry_<%=cid%> : resultMap_<%=cid%>.entrySet()) {
+		switch(entry_<%=cid%>.getKey()) {//for the old key
+		case org.talend.components.api.component.ComponentDefinition.RETURN_ERROR_MESSAGE :
```
gonchar-ivan: We can change constants values here
https://github.com/Talend/components/blob/master/components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java#L49-L55

to be "ERROR_MESSAGE", "NB_LINE", etc.

Then we won't need switch-case statement here.
Only 
container_<%=cid%>.setComponentData("<%=cid%>", entry_<%=cid%>.getKey(), entry_<%=cid%>.getValue())

wwang-talend: Yes, i know it, but please see the discussion in https://jira.talendforge.org/browse/TCOMP-165, Francis may have some other consideration.

gonchar-ivan: About ERROR_MESSAGE
Result class doesn't contain error message (and I think it should not contain it).
So container_<%=cid%>.setComponentData("<%=cid%>", "ERROR_MESSAGE", entry_<%=cid%>.getValue()) won't be called here.
Error message logic is common for all components. So I think error message should not appear in component, but should be handled in .javajet.
Check this jira for more details
https://jira.talendforge.org/browse/TUP-4825

Error message is set in TalendException.printStackTrace()
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.codegen/jet_stub/header.javajet#L455-L470

wwang-talend: you are right, will remove it,thanks.

gonchar-ivan: About ComponentDefinion constants. 
https://github.com/Talend/components/blob/master/components-api/src/main/java/org/talend/components/api/component/ComponentDefinition.java#L49-L55

Sorry, my bad. I found Francis comment about it in https://github.com/Talend/components/pull/152/files
I have understood why constant values were changed from uppercase to camel case notation.
Henceforth, return properties names (on components side) will follow camelCaseNotation.
That's why we can't use uppercase values (e.g. ERROR_MESSAGE). But for compatibility Studio requires uppercase values. And you have implemented mapping.

PR #567, opened 2016-07-15
TDI-36887:After do view documentation for job, the job/joblet node do

hwang-talend:
not show under documentation node.


ggu-talend:
The repository view should totally depend on the CNF with GeneratedDocContentProvider(Generated Node).
so should be some other problem of CNF trigger config for "org.eclipse.ui.navigator.navigatorContent" extension point.

PR #618, opened 2016-08-22
Bug fixes for TDI-37103, TDI-37108, TDI-37113

jlolling:
Empty log files will be deleted. This is not the actual solution for the missing log entries - this will rise an issue for EXASolution.
The target column was not defined if the source is a query
Pre defined connection causes a Exception


jzhao-talend:
Have been merged with amend.
https://jira.talendforge.org/browse/TDI-37103

PR #628, opened 2016-08-26
Feature/6.3/tdi 36698 instance roles iam account aws components

chmyga:
Added instance roles and Assume roles features to:
tAmazonEMRxxx
tAmazonRedshiftManage


wwang-talend:
the code is ok

PR #635, opened 2016-09-02
Feature/6.3.0 m1/export babili

nrousseau:


nrousseau:
can't be merged since there is the commit of irene for the M1 version, other commits are good and can be included manually

PR #643, opened 2016-09-09
TUP-5252: add JSON_TABLE controller which is a table with json input.

ycbai:
And fix the layout problem of Text, Combo, Checkbox, File, Table
controllers in the wizard at the same time.


nrousseau:
not sure the name json_table is really the best here... maybe dynamic could be better, not sure.
Still merged anyway, since it should be ok

PR #650, opened 2016-09-13
TDI-37285: fixed issues with getting out parameters from result.

chmyga:
Fixed the way we get parameters number to retrieve the value from call result.


wwang-talend:
a little difficult to review it 100%, if want to understand the fix, need to research it again. Do a simple check, it's ok.

PR #659, opened 2016-09-20
Feature/6.3/tcomp 230 flow variables retrieving

gonchar-ivan:

It can be merged only after following PR merged!
https://github.com/Talend/components/pull/239

### DIFF
--- a/main/plugins/org.talend.designer.codegen/jet_stub/generic/component_begin.javajet
+++ b/main/plugins/org.talend.designer.codegen/jet_stub/generic/component_begin.javajet
@@ -259,6 +259,23 @@ if (hasOutputOnly) {
 
         try {
             Object data_<%=cid%> = reader_<%=cid%>.getCurrent();
+            
+            // retrieving flow variables from reader
+            if (reader_<%=cid%> instanceof org.talend.components.api.component.runtime.HasReturnValues) {
+                java.util.Map<String, Object> flowVariables_<%=cid%> = reader_<%=cid%>.getFlowValues();
+                for(java.util.Map.Entry<String,Object> entry_<%=cid%> : flowVariables_<%=cid%>.entrySet()) {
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.codegen/jet_stub/generic/component_begin.javajet
```diff
@@ -259,6 +259,23 @@ if (hasOutputOnly) {
 
         try {
             Object data_<%=cid%> = reader_<%=cid%>.getCurrent();
+            
+            // retrieving flow variables from reader
+            if (reader_<%=cid%> instanceof org.talend.components.api.component.runtime.HasReturnValues) {
```
gonchar-ivan: I don't like this if condition appear in generated code, but as I know Reader or Writer instance is not available in Generator class.
Maybe I should add some method in ComponentDefinition to make this condition checked in Generator class?

main/plugins/org.talend.designer.codegen/jet_stub/generic/component_begin.javajet
```diff
@@ -259,6 +259,23 @@ if (hasOutputOnly) {
...
+                java.util.Map<String, Object> flowVariables_<%=cid%> = reader_<%=cid%>.getFlowValues();
+                for(java.util.Map.Entry<String,Object> entry_<%=cid%> : flowVariables_<%=cid%>.entrySet()) {
+                    StringBuilder studio_key_<%=cid%> = new StringBuilder();
+                    for (int i_<%=cid%> = 0; i_<%=cid%> < entry_<%=cid%>.getKey().length(); i_<%=cid%>++) {
```
gonchar-ivan: This loop is the same in 3 places: in begin part for Reader, in main part for Writer and in end part for after variables.
Maybe we could move this code to separate method?

gonchar-ivan:
The concept was changed, so this PR has no sense any more

PR #667, opened 2016-09-22
TCOMP-229 Implemented adding Iterate flow for TCOMP components

gonchar-ivan:
Related to https://jira.talendforge.org/browse/TCOMP-229

This PR adds "Iterate" connectors creation for TCOMP components.

After conversation with Francis, Sebastien and Ryan it was decided:
1. To not add notion of Iterate flow in TCOMP
2. Any component can have outgoing iterate connector
3. Only Standalone and Input components can have incoming  iterate connector


jzhao-talend:
As you said in the skype, we should use addStandardType(...)
And should not add 2 iterator line to the list.
As I understand, standalone and input components can have linke iterator input and output.
Other compoents only can link iterator output.
So 

>      if (hasInputIterateConnector) {
             //default allow link both input and output iterator
             addStandardType(listConnector, EConnectionType.ITERATE, parentNode);
        } else {
              //disable link input iterator
             addStandardType(listConnector, EConnectionType.ITERATE, parentNode).setMaxLinkInput(0);
       }

main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java
```diff
@@ -945,6 +951,8 @@ public int compare(Connector o1, Connector o2) {
             }
             addGenericType(listConnector, type, connector.getName(), parentNode, componentProperties, true);
         }
+        addGenericType(listConnector, EConnectionType.ITERATE, EConnectionType.ITERATE.getName(), parentNode, componentProperties, true);
```
jzhao-talend: No need add 2 iterator to the connector list.

jzhao-talend:
This changes is ok for me.

PR #677, opened 2016-09-27
Tdi 32687 enforce ntlm authentication

chmyga:


wwang-talend:
it's ok from the code

PR #678, opened 2016-09-27
TUP-15929: Cann't retrieve schema for HDFS

ycbai:


ggu-talend:
Try to use old delimted component to fix issue temp, later if the new delimited components are stable, will try to check and adapt to new one.

PR #740, opened 2016-10-24
TDI-37592 : Date Pattern doesn't take effect when set Type as 'Dynamic'

pyzhou-talend:
in schema of tFileInputExcel when reading .xls (97-2003) format.
https://jira.talendforge.org/browse/TDI-37592

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
@@ -402,22 +402,12 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
 			
 			String[] temp_row_<%= cid %> = new String[tempRowLength_<%=cid%>];
 			int actual_end_column_<%= cid %> = end_column_<%= cid %> >	row_<%= cid %>.length ? row_<%= cid %>.length : end_column_<%= cid %>;
-<%
-                            	if(hasDynamic){
-%>	
+
 				java.util.TimeZone zone = java.util.TimeZone.getTimeZone("GMT");
                 java.text.SimpleDateFormat sdf = new java.text.SimpleDateFormat(<%=datePattern%>);
                 sdf.setTimeZone(zone);
                 
-<%
-		                     	}else{
-%>
-				java.util.TimeZone zone = java.util.TimeZone.getTimeZone("GMT");
-				java.text.SimpleDateFormat sdf = new java.text.SimpleDateFormat();
-				sdf.setTimeZone(zone);
-<%
-		                     	}
-%>
+
 			for(int i=0;i<tempRowLength_<%=cid%>;i++){
 
 				if(i + start_column_<%= cid %> < actual_end_column_<%= cid %>){
@@ -432,28 +422,11 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
                             if(temp_row_<%=cid%>[i]!=null) {
                                 temp_row_<%=cid%>[i] = df_<%=cid %>.format(doubleCell_<%=cid%>);
                             }
-                     } else if(jxl.CellType.DATE== cell_<%=cid%>.getType()){
-<%
-                            	if(hasDynamic){
-%>
-	                            	if(i>=<%=dynamic_index-1%> && i<(<%=dynamic_index-1%> + dynamic_<%=cid%>.getColumnCount())){
+                    	} else if(jxl.CellType.DATE== cell_<%=cid%>.getType()){
+
 	                            		  	jxl.DateCell dc = (jxl.DateCell)cell_<%=cid%>;
                             				java.util.Date date = dc.getDate();
                         					temp_row_<%= cid %>[i] = sdf.format(date);
-	                            	}else{
-<%
-		                     	}
-%>
-										jxl.DateCell dc = (jxl.DateCell)cell_<%=cid%>;
-                            			java.util.Date date = dc.getDate();
-                        				temp_row_<%= cid %>[i] = sdf.format(date);
-<%
-                            	if(hasDynamic){
-%>
-                            		}
-<%
-                            	}
-%>        	
                         } else{
                             temp_row_<%= cid %>[i] = cell_<%=cid%>.getContents();
                         }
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
```diff
@@ -402,6 +402,22 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
 			
 			String[] temp_row_<%= cid %> = new String[tempRowLength_<%=cid%>];
 			int actual_end_column_<%= cid %> = end_column_<%= cid %> >	row_<%= cid %>.length ? row_<%= cid %>.length : end_column_<%= cid %>;
+<%
+                            	if(hasDynamic){
+%>	
+				java.util.TimeZone zone = java.util.TimeZone.getTimeZone("GMT");
```
chmyga: Are you sure we need to take the GMT timezone here? what if the timezone in the file is not GMT? That might lead to data inconsistency.

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
```diff
@@ -402,6 +402,22 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
...
+                java.text.SimpleDateFormat sdf = new java.text.SimpleDateFormat(<%=datePattern%>);
+                sdf.setTimeZone(zone);
+                
+<%
+		                     	}else{
+%>
+				java.util.TimeZone zone = java.util.TimeZone.getTimeZone("GMT");
+				java.text.SimpleDateFormat sdf = new java.text.SimpleDateFormat();
```
chmyga: Why don't we use the datePattern if !hasDynamic? Don't we need to use pattern here too?

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
```diff
@@ -416,6 +432,28 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
...
+                            				java.util.Date date = dc.getDate();
+                        					temp_row_<%= cid %>[i] = sdf.format(date);
+	                            	}else{
+<%
+		                     	}
+%>
+										jxl.DateCell dc = (jxl.DateCell)cell_<%=cid%>;
+                            			java.util.Date date = dc.getDate();
+                        				temp_row_<%= cid %>[i] = sdf.format(date);
+<%
```
chmyga: What is the difference between "if(i>=<%=dynamic_index-1%> && i<(<%=dynamic_index-1%> + dynamic_<%=cid%>.getColumnCount()))" and all other cases? the code looks totally similar. Are you sure we need that "if..else" clause?

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
```diff
@@ -416,6 +432,28 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
...
                                 temp_row_<%=cid%>[i] = df_<%=cid %>.format(doubleCell_<%=cid%>);
                             }
+                     } else if(jxl.CellType.DATE== cell_<%=cid%>.getType()){
+<%
+                            	if(hasDynamic){
+%>
+	                            	if(i>=<%=dynamic_index-1%> && i<(<%=dynamic_index-1%> + dynamic_<%=cid%>.getColumnCount())){
+	                            		  	jxl.DateCell dc = (jxl.DateCell)cell_<%=cid%>;
+                            				java.util.Date date = dc.getDate();
+                        					temp_row_<%= cid %>[i] = sdf.format(date);
```
jzhao-talend: I think here you can use FormatterUtils.format_Date(...) which is generic formatter utils

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
pyzhou-talend: Yes. I'm sure that we need to take the GMT timezone here. Because there is a little bug in jxl, it will adjust the date to the current timezone. So I can only change it back to get the origin date.

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
pyzhou-talend: if !hasDynamic the TOS will format the date according to pattern Automatically. I do this only to make sure our customers get the origin date. Even though they can't get the origin format, However wrong format is better than wrong data to me.

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
jzhao-talend: And about the pattern, we also need consider to add "Convert date column to string" for xls file.
Because different column can have different pattern. You can refer to excel 2007 which we have done for dynamic date format.

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
chmyga: Why do you need this if-else clause here then?

pyzhou-talend: Oh, you are right I don't need this "if..else" either.

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
jzhao-talend: ah~ Sorry. Forget my comment. They are some thing different.

pyzhou-talend:
I have removed the useless code.

main/plugins/org.talend.designer.components.localprovider/components/tFileInputExcel/tFileInputExcel_begin.javajet
```diff
@@ -402,6 +402,12 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
 			
 			String[] temp_row_<%= cid %> = new String[tempRowLength_<%=cid%>];
 			int actual_end_column_<%= cid %> = end_column_<%= cid %> >	row_<%= cid %>.length ? row_<%= cid %>.length : end_column_<%= cid %>;
+
+				java.util.TimeZone zone = java.util.TimeZone.getTimeZone("GMT");
```
jzhao-talend: It would be better add "_<%=cid%>" for "zone"/"sdf" to avoid variable duplicate issue when 2 tFileInputExcel are used in same sub-process

PR #768, opened 2016-11-08
TUP-16183 Fix: Component tRunJob has warnings of "The context name

kjwang-talend:
<value> of line <#> is not existed in column
https://jira.talendforge.org/browse/TUP-16183

### DIFF
--- a/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
+++ b/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
@@ -90,6 +90,7 @@ import org.talend.core.model.process.IProcess2;
 import org.talend.core.model.process.Problem;
 import org.talend.core.model.process.Problem.ProblemStatus;
 import org.talend.core.model.properties.JobletProcessItem;
+import org.talend.core.model.properties.ProcessItem;
 import org.talend.core.model.repository.ExternalNodesFactory;
 import org.talend.core.model.repository.IRepositoryViewObject;
 import org.talend.core.model.utils.NodeUtil;
@@ -113,6 +114,7 @@ import org.talend.core.ui.services.IDesignerCoreUIService;
 import org.talend.designer.core.CheckNodeManager;
 import org.talend.designer.core.DesignerPlugin;
 import org.talend.designer.core.ICheckNodesService;
+import org.talend.designer.core.IDesignerCoreService;
 import org.talend.designer.core.i18n.Messages;
 import org.talend.designer.core.model.components.AbstractBasicComponent;
 import org.talend.designer.core.model.components.EParameterName;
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
```diff
@@ -2964,6 +2971,24 @@ private void checkParameters() {
         }
 
     }
+    
+    private Object[] getContextParamsFromProcess() {
+        IElementParameter contextElemParam = this.getElementParameter(EParameterName.PROCESS_TYPE_CONTEXT.getName());
+        if (contextElemParam != null && this.process != null) {
+            String contextName = (String) contextElemParam.getValue();
```
ggu-talend: NOTE, this context name is for child job, not for current main job. so Shouldn't find it "process.getContextManager().getContext(contextName)" from current job.
pl, check "ProcessController.updateContextList", and will do same function. also PropertiesTableEditorView with "CONTEXT_PARAM_NAME_LIST"

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/properties/macrowidgets/tableeditor/PropertiesTableEditorView.java
```diff
@@ -261,18 +262,11 @@ public Object getCellEditorTypedValue(CellEditor cellEditor, Object originalType
                             return returnedValue;
                         };
                     });
-                    cellEditor.addListener(new ICellEditorListener() {
+                    
+                    ((CCombo) cellEditor.getControl()).addFocusListener(new FocusAdapter() {
```
ggu-talend: Change to use another listener, need check some operations, like press ENTER to apply value directly or such.

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
```diff
@@ -2964,6 +2974,50 @@ private void checkParameters() {
...
+                    String[] split = processId.split(";");
+                    processId = split[0];
+                }
+
+                ProcessItem processItem = ItemCacheManager.getProcessItem(processId, (String) jobVersionParam.getValue());
+                Process process = null;
+                if (processItem != null) {
+                    List<String> contextParameterNamesList = new ArrayList<String>();
+                    IDesignerCoreService service = CorePlugin.getDefault().getDesignerCoreService();
+                    process = (Process) service.getProcessFromItem(processItem);
```
ggu-talend: If only get the name list of context parameters from child job, I prefer no need load the child job to Process. can retrieve from Item directly. because when load to process, always have some performance issue.

PR #769, opened 2016-11-08
TDI-37735 : S3 Job throw errors but does complete successfully.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37735

tS3connection will only create the instance of connection object, but won't connect to S3. So tS3connection will always success. So I add this code to connect to S3.


pyzhou-talend:
I change the method to .getBucketLoggingConfiguration(String BucketName) sine it will only return a instance of BucketLoggingConfiguration (even if amazon don't record log for this bucket) which is quite tiny. And the only exception cause by this method is 404 NoSuchBucket which I will catch and ignore. If there is any other Exception, it proves that the connection doesn't succeed.

PR #773, opened 2016-11-09
TDI-37725 : tHttpRequest add error massage to global map

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37725

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tHttpRequest/tHttpRequest_main.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tHttpRequest/tHttpRequest_main.javajet
@@ -195,6 +195,7 @@ try{
             log.error("<%=cid%> - " + e.getMessage());
         <%}%>
         System.err.println(e.getMessage());
+        globalMap.put(currentComponent + "_ERROR_MESSAGE",e.getMessage());
     <%}else{%>
         throw e;
     <%}%>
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tHttpRequest/tHttpRequest_main.javajet
```diff
@@ -195,6 +195,7 @@ try{
             log.error("<%=cid%> - " + e.getMessage());
         <%}%>
         System.err.println(e.getMessage());
+        globalMap.put(currentComponent + "_ERROR_MESSAGE",e.getMessage());
```
jzhao-talend: We didn't use "currentComponent" in component side. 
In general, we use "<%=cid%>" instead of .

jzhao-talend:
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.components.localprovider/components/tHttpRequest/tHttpRequest_main.javajet#L154
And I think this one also need to change.

pyzhou-talend:
Thanks for pointing out my miss. I have made the change.

main/plugins/org.talend.designer.components.localprovider/components/tHttpRequest/tHttpRequest_main.javajet
pyzhou-talend: ok. I will make the change.

PR #774, opened 2016-11-09
Bugfix/6.3/tdi 36696 tmicrosoftcrminput 2015 paging

jzhao-talend:


chmyga:
Everything looks correct in the code.

PR #802, opened 2016-11-16
Bug/6.3.1/tdi 37712

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37712


main/plugins/org.talend.designer.components.localprovider/components/tFileInputPositional/tFileInputPositional_begin.javajet
```diff
@@ -601,6 +601,7 @@
             Object filename_<%=cid %> = <%=filename %>;
             java.io.BufferedReader in_<%=cid %> = null;
             org.talend.fileprocess.delimited.RowParser reader_<%=cid %> = null; 
+            org.talend.fileprocess.delimited.RowParserByByte byteReader_<%=cid %> = null;
```
jzhao-talend:
```
org.talend.fileprocess.delimited.RowParser reader_<%=cid %> = null; 
org.talend.fileprocess.delimited.RowParserByByte byteReader_<%=cid %> = null; 
```
A simple things, seems those 2 can't be use together. 
So better add condition for those code generation.
it would be let generate code clean.

PR #810, opened 2016-11-18
TDI-37454: Upgrade of tSalesforceInput -with Manual input of SOQL query-broken in 6.2.1

jzhao-talend:
https://jira.talendforge.org/browse/TDI-37454


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/constants/IGenericConstants.java
```diff
@@ -59,5 +59,12 @@
     public static final String IS_CONTEXT_MODE = "IS_CONTEXT_MODE"; //$NON-NLS-1$
 
     public static final String IS_DYNAMIC = "IS_DYNAMIC"; //$NON-NLS-1$
+    
+    /**
+     * Tag key estimates what would line separator in property would be replaced with.
+     * The property is which value can include line separator
+     */
+    public static final String LINE_SEPARATOR_REPLACED_TO = "LINE_SEPARATOR_REPLACED_TO";//$NON-NLS-1$
```
wwang-talend: the comment is not about the current commit : 
now we have to write the constant like this in two places, not good.

PR #815, opened 2016-11-23
TCOMP-331 Handle reject after close

francisu:


francisu:
Oops, this looks pretty bad.

PR #816, opened 2016-11-23
TCOMP-331 Support reject after close

francisu:


francisu:
This will need more work in code generation. In this case, of possible reject after close, the mains of the downstream components need to be generated here. I will update this soon.

francisu:
This has been updated to address the code generation issues fully and tested with the Snowflake components. It's ready to be merged if approved.

francisu:
Thanks @bchen-talend @gonchar-ivan for your fast attention to this!

PR #827, opened 2016-11-28
fix(TDI-36824): tDatasetOutput: Allow limit to be infinite if null (or empty) is set

wwang-talend:


wwang-talend:
use another solution, so close it

PR #830, opened 2016-11-29
fix(TDI-36824): tDatasetOutput: Allow limit to be infinite if null (or empty) is set

wwang-talend:


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java
```diff
@@ -197,7 +197,8 @@ private String getTranslatedValue(final String nameValue) {
         for (Property<?> child : componentDefinition.getReturnProperties()) {
             nodeRet = new NodeReturn();
             nodeRet.setType(ComponentsUtils.getTalendTypeFromProperty(child).getId());
-            nodeRet.setDisplayName(ComponentReturnVariableUtils.getTranslationForVariable(child.getName(), child.getDisplayName()));
+            nodeRet.setDisplayName(
+                    ComponentReturnVariableUtils.getTranslationForVariable(child.getName(), child.getDisplayName()));
```
sgandon: strange formatting.

PR #831, opened 2016-11-29
TUP-16466:XML/XSD:Studio hang on after detect update

hwang-talend:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/update/cmd/UpdateNodeParameterCommand.java
```diff
@@ -311,7 +311,8 @@ private void updateProperty() {
                                     || param.getFieldType() == EParameterFieldType.MEMO_SQL) {
                                 continue;
                             }
-                            if (param.getFieldType().equals(EParameterFieldType.FILE) && isXsdPath) {
+                            //    TUP-16466:XML/XSD:Studio hang on after detect update
+                            if (node.getComponent().getName().startsWith("tESB")&&param.getFieldType().equals(EParameterFieldType.FILE) && isXsdPath) {
```
zwxue: why judging tESB here?

PR #832, opened 2016-11-29
[TESB-18057] Fix

vbogatyrov:
When a job is undeployed in Runtime, corresponding thread receives interruption request from Karaf, and job controller treats the request as an indicator that the job is undeployed,  otherwise it tries to restart the job (probably as a sort of failure recovery).

When job multi thread execution is turned on, corresponding interruption request is sent to one of the subthreads, but not to the thread that is monitored by job controller, which leads to infnite attempts to restart the job and unlimited growth of logs.

Proposed fix sends interruption request to the thread, monitored by job controller in case if one of the subthreads receives interruption request.

### DIFF
--- a/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
+++ b/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
@@ -749,7 +749,7 @@ this.globalResumeTicket = false;//to run others jobs
                         status = localStatus;
                     }
 
-                    if ("true".equals(((java.util.Map) threadLocal.get()).get("ESBJobInterruptedException"))) {
+                    if ("true".equals(((java.util.Map) threadLocal.get()).get("JobInterrupted"))) {
                         launchingThread.interrupt();
                     }
 
--- a/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
+++ b/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
@@ -768,6 +768,8 @@ this.globalResumeTicket = false;//to run others jobs
             Thread.sleep(10);
         } catch (java.lang.InterruptedException e) {
             interrupted = true;
+        } catch (java.lang.Exception e) {
+            e.printStackTrace();
         }
     }
 
--- a/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
+++ b/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
@@ -707,6 +707,8 @@ this.globalResumeTicket = true;//to run tPreJob
 
 this.globalResumeTicket = false;//to run others jobs
 
+final Thread launchingThread = Thread.currentThread();
+
 <%
 
     //5. all others sub-job (MultiThread mode)
@@ -748,6 +750,10 @@ this.globalResumeTicket = false;//to run others jobs
                         status = localStatus;
                     }
 
+                    if ("true".equals(((java.util.Map) threadLocal.get()).get("ESBJobInterruptedException"))) {
+                        launchingThread.interrupt();
+                    }
+
                     runningThreadCount.add(-1);
                 }
             }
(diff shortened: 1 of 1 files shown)


vbogatyrov:
The issue needs to be fixed on master, 6.1 and 6.2

main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
```diff
@@ -757,13 +763,20 @@ this.globalResumeTicket = false;//to run others jobs
...
-            e.printStackTrace();
+        } catch (java.lang.InterruptedException e) {
+            interrupted = true;
         }
     }
+
+    if (interrupted) {
+        Thread.currentThread().interrupt();
+    }
+
```
chmyga: Are you trying to stop the job processing? what about closing some of the resources or statuses processing? sometimes, if you don't close the channel, the source might be unavailable after that - some lock might be put over the source and we won't have any access to it at all. Or are you trying to do something else? As interrupt doesn't look like it's going to stop the thread right at the moment of the call.

main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
```diff
@@ -707,6 +707,8 @@ this.globalResumeTicket = true;//to run tPreJob
 
 this.globalResumeTicket = false;//to run others jobs
 
+final Thread launchingThread = Thread.currentThread();
+
```
vbogatyrov: Done

main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
vbogatyrov: The scenario is the following: job with multithreading turned on is deployed into TESB Runtime. Then the user undeploys the job using Karaf command kar:undeploy which will cause return from 
https://github.com/Talend/tesb-rt-se/blob/master/job/controller/src/main/java/org/talend/esb/job/controller/internal/RuntimeESBProviderCallback.java#L120

In case of non-multithreaded job Karaf also interrupts job thread that is checked by https://github.com/Talend/tesb-rt-se/blob/master/job/controller/src/main/java/org/talend/esb/job/controller/internal/RuntimeESBProviderCallback.java#L116 , which makes infinite while loop finish and job stops.

In case of multi-threaded job, one of the subthreads creted at https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet#L731 receives interruption from Karaf, but not the main job thread, which leads to inifinte attempts to restart job by Job controller.

By calling https://github.com/Talend/tdi-studio-se/pull/832/commits/96465d3e42d1b844bea9fea272fa7338103ee5e1#diff-c0cd44ca5e3d8407d1c950cf91285c59R776 the interruption is made to thread that is actually monitored by Job controller which stops the while loop and finishes the job.

main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
```diff
@@ -757,13 +762,20 @@ this.globalResumeTicket = false;//to run others jobs
...
 %>
+    boolean interrupted = false;
     while (runningThreadCount.getCount() > 0) {
         try {
             Thread.sleep(10);
-        } catch (java.lang.Exception e) {
-            e.printStackTrace();
+        } catch (java.lang.InterruptedException e) {
+            interrupted = true;
         }
```
wwang-talend: better to not remove the exception catcher for the old action

main/plugins/org.talend.designer.codegen/jet_stub/footer.javajet
```diff
@@ -748,6 +749,10 @@ this.globalResumeTicket = false;//to run others jobs
                         status = localStatus;
                     }
 
+                    if ("true".equals(((java.util.Map) threadLocal.get()).get("JobInterrupted"))) {
```
wwang-talend: when will you set the status code "JobInterrupted" in the subjob?

vbogatyrov: It is set here https://github.com/Talend/tesb-studio-se/blob/master/main/plugins/org.talend.designer.esb.components.ws.provider/components/tESBProviderRequestLoop/tESBProviderRequestLoop_end.javajet#L22

wwang-talend:
LGTM, the logic is like this now : 
if some sub thread receive some request("JobInterrupted" status is true) for interrupting the hold status, it should pass the interrupt status to parent thread. And the parent thread will set the interrupt status to true when all the sub job is over(finished or exception) if any sub job have send the interrupt request to it.

pteyssier:
Hello, code too dangerous to be merged now. We will integrate after the release branch 6.3.1 is created.

PR #847, opened 2016-12-01
TBD-4268: Ensure that classpaths are consistent for jobs run in Studi…

RyanSkraba:
Ensure that classpaths are consistent for jobs run in Studio and local jobs.

Currently, the order of dependencies in the generated classpath is undefined.  In practice, it happens to be consistent between runs of the same job (with one *exception*), but this is not guaranteed anywhere in the code.

The *exception* is that a job run from the Studio from the Run tab in DI does not use the same classpath order as a job exported and run from the launcher script (`_run.sh`).  We've encountered a problem with a job that works in one case but not the other.

When exporting a job, the order of the dependencies comes from iterating over a `Set<>` (which is undefined).

When running in the Studio, the order of the dependencies comes from the list of jar files in the `./lib` directory (where `java.io.File.listFiles()` also returns an undefined order).

This PR uses the existing `ModulesNeededComparator` to force an order on the jar files in the classpath, but there's probably a better deterministic order.  This solves our problem.

This might introduce regressions in jobs that relied on the previously undefined order of the classpath.

### DIFF
--- a/main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/java/JavaProcessor.java
+++ b/main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/java/JavaProcessor.java
@@ -129,6 +129,7 @@ import org.talend.designer.core.model.utils.emf.talendfile.ProcessType;
 import org.talend.designer.core.ui.editor.CodeEditorFactory;
 import org.talend.designer.core.ui.editor.nodes.Node;
 import org.talend.designer.core.ui.editor.process.Process;
+import org.talend.designer.core.utils.ModuleNeededComparator;
 import org.talend.designer.maven.utils.PomUtil;
 import org.talend.designer.runprocess.ItemCacheManager;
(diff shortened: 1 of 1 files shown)


RyanSkraba:
I've been thinking -- this probably *won't* cause regressions...

When exporting a job or running on a job server, the order of dependencies depends on the hashCode() of the ModuleNeeded instances(since it's being read from a Set).  Since we don't control the hash code in any way, it's more or less random whether one dependency appears before another.

If there are N Talend jobs in the world that have dependencies with class collisions (i.e. depend on a specific ordering of dependencies to work), we should expect about half of them to fail with the current logic.

The same logic applies to jobs that run in the Studio -- since the order is also undefined, but different, we would only expect 1/4 of the N jobs to work both from the Studio AND from an exported job.  (Probably not quite that bad, because I'd suspect that listFiles() is based on something other than a good hashCode())

Since we don't see a lot of problems with class collisions, N must be pretty close to zero... TBD-4268 is the only time I've encountered this problem (Salesforce doesn't work with a specific version of Hive).

PR #849, opened 2016-12-02
TUP-16510:tRedshiftOutput Data Viewer fails.

hcyi:


hcyi:
After checked , will use another way to fix this problem , close this pull request .

PR #852, opened 2016-12-05
TDI-37774 : removed log4j

pyzhou-talend:


pyzhou-talend:
There is no log4j include in version 5.5. So I removed log4j.

PR #866, opened 2016-12-13
TDI-37961: tMysqlnput with mariadb driver cause compilation error

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37961


jzhao-talend:
I also have a question why need a migration for this jira ?
I have a situation:
1) create a tMysqlInput with db version mysql 5/4 then check the "Enable streaming"
2) Change db version to MariaDB
The compiler error also occur.
So the migration is not necessary
Should avoid the issue in javajet. the code caused compiler error should generated when MariaDB unchecked and "Enable streaming" checked

PR #868, opened 2016-12-13
feat(TCOMP-230): add Reader wrapping with FlowVariablesReader

gonchar-ivan:

It shows how FlowVariablesReader is supposed to be used in codegen.
Related PR is https://github.com/Talend/tcommon-studio-se/pull/794

Do not merge right now. It is likely to be changed according how FlowVariablesReader will be packed


gonchar-ivan:
It is outdated.
This change was included in following PR
https://github.com/Talend/tdi-studio-se/pull/886

PR #870, opened 2016-12-14
TDI-37961 :  tMysqlnput with mariadb driver cause compilation error

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37961


pteyssier:
Duplicate.

PR #880, opened 2016-12-16
TESB-17533: Expose attributes as JSON elements

sebhoss:


csuntalendbj:
reviewed , and tested on 6.5.0 snapshot

PR #905, opened 2016-12-29
TUP-16639:Impossible to filter components from the new framework

hwang-talend:


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java
```diff
@@ -116,6 +117,12 @@
     private ComponentDefinition componentDefinition;
 
     private RuntimeInfo runtimeInfo;
+    
+    private Boolean visible = null;
+    
+    private Boolean technical = null;
+    
+    private Boolean visibleFromComponentDefinition = null;
```
ggu-talend: is it possible to extract the fields into parent class AbstractBasicComponent.

main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/utils/ComponentsUtils.java
```diff
@@ -133,7 +135,33 @@ public static void loadComponents(ComponentService service) {
...
+                }
                 componentsList.add(currentComponent);
+                // if the component is not needed in the current branding,
+                // and that this one IS a specific component for code generation,
+                // hide it
+                if (hiddenComponent
+                        && (currentComponent.getOriginalFamilyName().contains("Technical") || currentComponent.isTechnical())) {
+                    currentComponent.setVisible(false);
+                    currentComponent.setTechnical(true);
+                }
```
ggu-talend: It's duplicate codes with ComponentsFactory line 529 - 555, is it possible to extract to parent class AbstractBasicComponent to have new method for example "isAvailable", then call it:
if (component.isAvailable){
//add/put into list/map
}else{
continue;
}

PR #916, opened 2017-01-05
feat(TDI-35748):salesforce connection save and reuse session

jzhao-talend:
* migration from old framework
* add directory mapping with widget type


jzhao-talend:
Merge together with Talend/components/pull/403

PR #925, opened 2017-01-10
TUP-16584：Job run Tab >> Advanced Settings >> Add JVM argument - No

hwang-talend:
escape character works [MacOS]

### DIFF
--- a/main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/utils/JobVMArgumentsUtil.java
+++ b/main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/utils/JobVMArgumentsUtil.java
@@ -1,0 +1,80 @@
+// ============================================================================
+//
+// Copyright (C) 2006-2016 Talend Inc. - www.talend.com
+//
+// This source code is available under agreement available at
+// %InstallDIR%\features\org.talend.rcp.branding.%PRODUCTNAME%\%PRODUCTNAME%license.txt
+//
+// You should have received a copy of the agreement
+// along with this program; if not, write to Talend SA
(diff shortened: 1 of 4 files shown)


main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/utils/JobVMArgumentsUtil.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    private static final List<String> EMPTY_STRING_LIST = Collections.unmodifiableList(new ArrayList<String>());
+    
+    public List<String> readString(String stringList) {
+        if (stringList == null || "".equals(stringList)) { //$NON-NLS-1$        
+            return EMPTY_STRING_LIST;
+        }
+        ArrayList<String> result = new ArrayList<String>(50);
+        if(!isJson(stringList)){
+            return result;
```
ggu-talend: Means, there is no migration task? but non-json, will return empty list always? or need do it like before to split via space for list?

main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/utils/JobVMArgumentsUtil.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        try {
+            JSONObject root = new JSONObject(stringList);
+            Object obj =  root.get("JOB_RUN_VM_ARGUMENTS");//$NON-NLS-1$
```
ggu-talend: BTW, why need one key in json, if only save for the jvm args list, just JSONArray will be enough?

PR #932, opened 2017-01-15
BatchUpdateExceptionHelper used to setup a meaningful exception in the

jlolling:


pyzhou-talend:
Sorry for the delays,kind of busy these days. We are more than grateful for this pull request.  I have to modified a bit about this PR. I will try to finish that and merge the PR this week(or next).

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-28181

main/plugins/org.talend.designer.components.localprovider/components/tEXAOutput/tEXAOutput_end.javajet
```diff
@@ -115,9 +118,7 @@ imports="
                 deletedCount_<%=cid%> += countSum_<%=cid%>;
             <%
             }
-            dbLog.logPrintedException("e.getMessage()");%>
-            System.err.println(e.getMessage());
```
wwang-talend: should keep it as it's a function for "die on error".

main/plugins/org.talend.designer.components.localprovider/components/tOracleOutput/tOracleOutput_main.javajet
```diff
@@ -1245,17 +1248,20 @@ skeleton="../templates/db_output_bulk.skeleton"
                             <%dbLog.batch().executeTry(dbLog.str(dataAction));%>
                             pstmt_<%=cid %>.executeBatch();
                             <%dbLog.batch().executeDone(dbLog.str(dataAction));%>
-                        }catch (java.sql.BatchUpdateException e_<%=cid%>){
+                        }catch (java.sql.BatchUpdateException e){
                             <%
                             if(("true").equals(dieOnError)) {
                             %>
-                                throw(e_<%=cid%>);
+
```
wwang-talend: a little strange at this part as it print the message at once when "die on error", will check it

main/plugins/org.talend.designer.components.localprovider/components/templates/DB/Output/BatchUpdateExceptionHelper.javajet
```diff
@@ -0,0 +1,24 @@
...
+                fullErrorMessage.append("Error Code: ").append(((java.sql.SQLException)e).getErrorCode()).append('\n');
+                fullErrorMessage.append("Message: ").append(e.getMessage());
+                Throwable t = ex.getCause();
+                while(t != null) {
+                    fullErrorMessage .append("Cause: ").append(t.toString());
+                    t = t.getCause();
+                }
+            }
+        }
+        System.err.println(fullErrorMessage);
```
wwang-talend: this is a side effect for this method, but logic is OK. And system.err some details information and throw exception with some other information at once, a little strange in my view. But after checking the code, look the logic is ok.

jlolling:
It looks like you mixe up the getClause() method with the getNextException() method!
Wew need getNextException and not getClause()!!!

jlolling:
You have caused a serious problem!

You use the method getCause() to get the Exception chain but this is not correct! You have to use the method getNextException() !

Please revert your changes and take a look at my original pull request!

Best regards
Jan

pyzhou-talend:
I get the code from this page https://docs.oracle.com/javase/tutorial/jdbc/basics/sqlexception.html
which came from oracle. I think to literate Throwable from the SQLException have the same effect from getNextException().

jlolling:
Hi,

NO this is not correct! The cause of an exception is what you have given as root cause. This is very different to what we need in thr BatchException.
By the way, the causes will be printed already printed out by all logging frameworks. 

Please use the getNextException!!!!!!!!!!!!!!!!!!

Best regards
Jan Lolling

jlolling:
Sorry,

It is also not a good idea to iterate through all getNextExceptions because it depends on the database vendor, if they collect all problems and reject the batch or reject the batch just at the first problem.
You can run into a situation, where you have thausends of exceptions if you are trying to get ALL getNextExceptions. That is also the reason why I have suggested to get only the first one!

The code suggested from Oracle is WRONG in the aspect of dealing with batch exceptions - for sure we cannot use the suggested method for all other vendors!

Please belief me, I have 20 years experiences in dealing with that!

Best regards
Jan

jlolling:
You have perhaps missed this statement from Oracle:

A reference to any chained exceptions. If more than one error occurs, the exceptions are referenced through this chain. Retrieve these exceptions by calling the method SQLException.getNextException on the exception that was thrown.

Best regards
Jan

pyzhou-talend:
Hi,

Thanks for the explaining. I will revert my modification and do as you said.

Best regards
pyzhou

jlolling:
Hi pyzhou,

no problem, I am happy you are not restistent against consulting ;-)

Cheers
Jan

pyzhou-talend:
I create another PR to revert this and redo as you said. you can check it here https://github.com/Talend/tdi-studio-se/pull/1223 .

PR #935, opened 2017-01-16
TMDM-10272 tDataStewardshipTaskOutput: Comments are not evaluated for each row but only once at the begining

htyin:
Some component property need set value in component_main.javajet


main/plugins/org.talend.designer.codegen/jet_stub/generic/component_main.javajet
```diff
@@ -134,6 +143,88 @@ if(hasInput){
             // connections.
 
             %>
+                                  
+            <%
+			List<Component.CodegenPropInfo> propsToProcess = component.getCodegenPropInfos(componentProps);
```
jzhao-talend: Most of this part of added codes are duplicated with begin part. Do you find a way to reuse them ?
like what we have done in `component_util_indexedrecord_to_rowstruct.javajet`

main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/constants/IGenericConstants.java
```diff
@@ -66,5 +66,5 @@
      */
     public static final String LINE_SEPARATOR_REPLACED_TO = "LINE_SEPARATOR_REPLACED_TO";//$NON-NLS-1$
 
-
+    public static final String SET_VALUE_IN_RUNTIME = "SET_VALUE_IN_RUNTIME";//$NON-NLS-1$
```
jzhao-talend: About this key of tagged value, you'd better confirm with TCOMP team Ivan or Seb.

jzhao-talend:
@gonchar-ivan

gonchar-ivan:
Performance aspect:

1) After each record processed this code tries to reset properties. This will hugely decrease job execution. Checking SET_VALUE_IN_RUNTIME key is a good idea (this is to reset only those properties, which allows dynamic values. Other properties will be skipped). But this check also will decrease performance (I'm not sure how much, but this is critical topic).

Components aspect:
1) This change assumes that Properties class is available in runtime classes (SourceOrSink, Reader, Writer etc) as a field. This is not a rule. Properties is passed to SourceOrSink.initialize() method. It this method Properties may be saved as field or may be saved as primitive values only. Even more latter is preferable (to reduce size of serialized SourceOrSink). So, after initialize() method Properties class might not be used anymore.
(ok, this code won't break anything for this case... but it will try to reset properties idle) 
2) Not sure whether setting value as "row1.someColumn" will work for output of TCOMP components. It should, because IndexedRecords are converted to row after TCOMP components... but I'm not sure.
Have you tried it with new components?

Comment:
Recently I added codegen-utils.jar library. This lib is available for generated code. 
It would be great to move as more as possible code from .javajet(s) to that lib.
Adv:
1. We may have tests for that library (which is not possible for .javajet)
2. Code will be more clean and readable

Conclusion:
1. This should work
2. We need better solution for this
3. And I'm not sure TCOMP team will be able to solve this in the near future

lpw82gf:
@gonchar-ivan  Does it mean we can merge this PR or not? If not, what further work need to do from our side?

gonchar-ivan:
After discussion with Pierre it can't be merged.

PR #936, opened 2017-01-17
fix(TDI-37987): tMicrosoftCrmOutput has NullPointerException

chmyga:
* Add "Ignore null" property

* Add ignoring null values code

* Add fix for NPE when set primitive value from null(Online 2011 API)

### DIFF
--- a/main/plugins/org.talend.repository/plugin.xml
+++ b/main/plugins/org.talend.repository/plugin.xml
@@ -2823,11 +2823,11 @@
         </projecttask>
         <projecttask
               beforeLogon="false"
-              breaks="5.6.2"
+              breaks="6.2.1"
               class="org.talend.repository.model.migration.AddNewIgnoreNullSettingForTMicrosoftCrmOutputTask"
               id="org.talend.repository.model.migration.AddNewIgnoreNullSettingForTMicrosoftCrmOutputTask"
               name="AddNewIgnoreNullSettingForTMicrosoftCrmOutputTask"
-              version="6.0.0">
+              version="6.2.2">
         </projecttask>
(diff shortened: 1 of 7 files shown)


main/plugins/org.talend.repository/plugin.xml
```diff
@@ -2821,6 +2821,14 @@
               name="FixUnevenItemContextParametersMigrationTask"
               version="6.2.2">
         </projecttask>
+        <projecttask
+              beforeLogon="false"
+              breaks="5.6.2"
+              class="org.talend.repository.model.migration.AddNewIgnoreNullSettingForTMicrosoftCrmOutputTask"
+              id="org.talend.repository.model.migration.AddNewIgnoreNullSettingForTMicrosoftCrmOutputTask"
+              name="AddNewIgnoreNullSettingForTMicrosoftCrmOutputTask"
+              version="6.0.0">
```
jzhao-talend: https://jira.talendforge.org/browse/TDI-21417
Based on the comment which add by  Seb to  explain  "version".
The version should be 6.2.2 in this branch 6.2

main/plugins/org.talend.repository/plugin.xml
```diff
@@ -2821,6 +2821,14 @@
               name="FixUnevenItemContextParametersMigrationTask"
               version="6.2.2">
         </projecttask>
+        <projecttask
+              beforeLogon="false"
+              breaks="5.6.2"
```
jzhao-talend: "breaks" set to 6.2.1 would be better.

chmyga:
changed the migration breaks/version

PR #940, opened 2017-01-19
TDI-37833:tEXAOutput component does not support dynamic schema

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37833

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tEXAOutput/tEXAOutput_main.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tEXAOutput/tEXAOutput_main.javajet
@@ -143,10 +143,10 @@ skeleton="@{org.talend.designer.components.localprovider}/components/templates/d
             IConnection conn = conns_dynamic.get(0);
             if(!("".equals(insertColName.toString()))) {
 %>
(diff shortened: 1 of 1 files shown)


pyzhou-talend:
For Exasol, the dot(.) of schema in double quotation will lead to Exception.

main/plugins/org.talend.designer.components.localprovider/components/tEXAOutput/tEXAOutput_main.javajet
```diff
@@ -143,10 +143,10 @@ skeleton="@{org.talend.designer.components.localprovider}/components/templates/d
             IConnection conn = conns_dynamic.get(0);
             if(!("".equals(insertColName.toString()))) {
 %>
-                String insert_<%=cid%> = "INSERT INTO \"" + tableName_<%=cid%> + "\" (<%=insertColName.toString()%>, "+DynamicUtils.getInsertIntoStmtColumnsList(<%=conn.getName()%>.<%=getDynamicColumn()%>, "<%=dbmsId %>")+") VALUES (<%=insertValueStmt.toString()%>, "+DynamicUtils.getInsertIntoStmtValuesList(<%=conn.getName()%>.<%=getDynamicColumn()%>)+")";
+                String insert_<%=cid%> = "INSERT INTO " + tableName_<%=cid%> + " (<%=insertColName.toString()%>, "+DynamicUtils.getInsertIntoStmtColumnsList(<%=conn.getName()%>.<%=getDynamicColumn()%>, "<%=dbmsId %>")+") VALUES (<%=insertValueStmt.toString()%>, "+DynamicUtils.getInsertIntoStmtValuesList(<%=conn.getName()%>.<%=getDynamicColumn()%>)+")";
```
wwang-talend: The change above is OK as :
1. it only make the dynamic action the same with nondynamic action which no double quote for tablename too.
2. After make sure with Pengyu, if double quote exists for tablename, it will throw some exception when tablename is like "myschema.mytable", or even when set the schema in the component db schema setting and set tablename like "mytable".

PR #955, opened 2017-02-01
Patch/6.3.1/new_functionality_SSKMS

marciuc:
Add new tS3Put component to support Server side encryption (SSE) with AWS Key management system (KMS).
[tS3PutSSKMS.zip](https://github.com/Talend/tdi-studio-se/files/745072/tS3PutSSKMS.zip)


pteyssier:
That is not the way to make PR in github.

PR #964, opened 2017-02-07
TDI-38166

mrizzato:
Hello,
I have patched the attached file to fix a problem when selecting the Delete Emails From Server in tPOP component using imap protocol, as reported in issue [JIRA] (TDI-38166).

[tPOP_begin.javajet.zip](https://github.com/Talend/tdi-studio-se/files/757051/tPOP_begin.javajet.zip)


pteyssier:
This PR is not possible to integrate as-is, it's based on 2 different branches, could you regenerate it to the equivalent branch please (maintenance/6.3)?

mrizzato:
Hello,

Done as requested.

PR #969, opened 2017-02-08
TDI-37755:Using the origin replaceCRLFInMEMO_SQL method.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37755


pyzhou-talend:
"/r" is only used by old version of Mac. We can omit it and sothat the origin replaceCRLFInMEMO_SQL  could be reserved.

PR #974, opened 2017-02-10
TDI-38177:tFileDelimitedOutput compile error if split every is empty.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38177

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tFileOutputDelimited/tFileOutputDelimited_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tFileOutputDelimited/tFileOutputDelimited_begin.javajet
@@ -381,7 +381,13 @@ if(("false").equals(ElementParameterParser.getValue(node,"__CSV_OPTION__"))) {
             }
             %>
             int nb_line_<%=cid%> = 0;
+            <%
+            if(split){
+            %>
             int splitEvery_<%=cid%> = <%=splitEvery %>;
+            <%
+            }
+            %>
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tFileOutputDelimited/tFileOutputDelimited_begin.javajet
```diff
@@ -381,7 +381,13 @@ if(("false").equals(ElementParameterParser.getValue(node,"__CSV_OPTION__"))) {
             }
             %>
             int nb_line_<%=cid%> = 0;
+            <%
```
jzhao-talend: For those 2 input mode, some code is duplicate between them.
is it possible, move them out of 2 input mode. and put them public ?
Then it would be easy to be maintained.

pyzhou-talend: ok. I will try.

jzhao-talend: or you can ask Pierre whether my suggestion is necessary.
Because this components would be move to new framework in future.

pyzhou-talend: I've confirmed with Pierre. He said that this component is too impotent. And we will keep the duplication.

PR #980, opened 2017-02-15
TUP-17093:[Talend Studio] Compilation error not shown on a box

hwang-talend:


nrousseau:
please try to keep the same format as before for the commit.
It's impossible to understand what was done here.

PR #540, opened 2016-06-29
bugfix/6.2.2/TUP-4925 : The code of method

pxmiao-talend:
tMSSqlInput_3Process(Map<String,Object>) is exceeding the 65535 bytes
limit
https://jira.talendforge.org/browse/TUP-4925

PR #981, opened 2017-02-15
TDI-38197: Performance issue for tMssqlOutput.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38197

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tMSSqlOutput/tMSSqlOutput_main.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tMSSqlOutput/tMSSqlOutput_main.javajet
@@ -1556,32 +1556,16 @@ if((cid.equals("talendLogs_DB") || cid.equals("talendStats_DB") || cid.equals("t
         	%>
             	//////////batch execute by batch size///////
             	class LimitBytesHelper_<%=cid%>{
-            		public void limitBytePart1(int batchSize_<%=cid%>,int batchSizeCounter_<%=cid%>,int insertedCount_<%=cid%>,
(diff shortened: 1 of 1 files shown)


pyzhou-talend:
The inner class disable the branch clean. So after first batch, it will commit on every insert. And here's another issue, it will create 10,000 instance of the inner class.

main/plugins/org.talend.designer.components.localprovider/components/tMSSqlOutput/tMSSqlOutput_main.javajet
```diff
@@ -1558,7 +1558,6 @@ if((cid.equals("talendLogs_DB") || cid.equals("talendStats_DB") || cid.equals("t
             	class LimitBytesHelper_<%=cid%>{
             		public void limitBytePart1(int batchSize_<%=cid%>,int batchSizeCounter_<%=cid%>,int insertedCount_<%=cid%>,
```
jzhao-talend: A question not about your modification: Could you check the NB line value about this component?
I just think they are invalid when commit split into batches.

jzhao-talend:
@pteyssier Because of this issue is caused by #540
I found another issue about this PR. So I prefer to recheck #540 and revert wrong part.
(changes about batch part)

main/plugins/org.talend.designer.components.localprovider/components/tMSSqlOutput/tMSSqlOutput_main.javajet
jzhao-talend: Confirmed with simple test. The nb line would be invalid, if it included several batches.

PR #983, opened 2017-02-15
fix(TCOMP-230): fix NPE problem in case of empty Iterable returned fr…

gonchar-ivan:
…om wrapped Writer

Fixes NPE problem

### DIFF
--- a/main/plugins/org.talend.designer.components.libs/libs_src/talend-codegen-utils/src/main/java/org/talend/codegen/flowvariables/runtime/FlowVariablesWriter.java
+++ b/main/plugins/org.talend.designer.components.libs/libs_src/talend-codegen-utils/src/main/java/org/talend/codegen/flowvariables/runtime/FlowVariablesWriter.java
@@ -137,6 +137,8 @@ public class FlowVariablesWriter<WriteT> implements WriterWithFeedback<WriteT, O
                     successDataProcessor = new MainDataProcessor();
                 }
                 firstSuccessData = false;
+            } else {
+                return successfulWrites;
             }
         }
         return successDataProcessor.processDataIterable(successfulWrites);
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.designer.components.libs/libs_src/talend-codegen-utils/src/main/java/org/talend/codegen/flowvariables/runtime/FlowVariablesWriter.java
```diff
@@ -137,6 +137,8 @@ public WriteT close() throws IOException {
                     successDataProcessor = new MainDataProcessor();
                 }
                 firstSuccessData = false;
+            } else {
+                return successfulWrites;
             }
         }
         return successDataProcessor.processDataIterable(successfulWrites);
```
wwang-talend: seems that successDataProcessor may be null, please check it

main/plugins/org.talend.designer.components.libs/libs_src/talend-codegen-utils/src/main/java/org/talend/codegen/flowvariables/runtime/FlowVariablesWriter.java
```diff
@@ -182,6 +184,8 @@ public WriteT close() throws IOException {
                     rejectDataProcessor = new MainDataProcessor();
                 }
                 firstSuccessData = false;
+            } else {
+                return rejectedWrites;
             }
         }
         return rejectDataProcessor.processDataIterable(rejectedWrites);
```
wwang-talend: the same with successDataProcessor

gonchar-ivan:
thanks

PR #988, opened 2017-02-16
fix(TDI-38232): fix index out of bounds exception in the tfileoutputmsxml

akhabali:
this is for [https://jira.talendforge.org/browse/TDI-38232 ](https://jira.talendforge.org/browse/TDI-38232 )

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tFileOutputMSXML/tFileOutputMSXML_begin.inc.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tFileOutputMSXML/tFileOutputMSXML_begin.inc.javajet
@@ -409,22 +409,28 @@ class GenerateToolByDom4j{
 %>
 			}
 			
+			//IF the element is at the end of the xml branch we just append it to the end of the elements list
 			if(order_<%=cid%> == allElements_<%=currEleName%>_<%=cid%>.size()) {
 				orderHelper_<%=cid%>.fillTailAppendMap("<%=tool.connName + "_" + currEleName%>",true);
 			}
 			
 			orderHelper_<%=cid%>.setOrder("<%=tool.connName + "_" + currEleName%>",order_<%=cid%>);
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tFileOutputMSXML/tFileOutputMSXML_begin.inc.javajet
```diff
@@ -415,9 +415,6 @@ class GenerateToolByDom4j{
 			
 			orderHelper_<%=cid%>.setOrder("<%=tool.connName + "_" + currEleName%>",order_<%=cid%>);
 			orderHelper_<%=cid%>.fillRecalculateOrderMap("<%=tool.connName + "_" + currEleName%>",false);
-		} else {
```
wwang-talend: if remove the code, the order of loop element will be wrong.

wwang-talend: case
please consider this case, you will the loop order is different after the fix.

akhabali: You are right. I didn't know that wee should keep the order of loop elements.
but i think that the incremented order exceed the number of the elements created in the xml tree for some reason.

i prepare an other fix for that by considering the loop order.

akhabali:
**The calculated order is the index of the element in the XML Branch. we can't just increment it.**
this is the case that break the existing logic.
when we have this kind of structure : 
* Root (of the XML)
  * Goup
    * Loop (_the order is 0 in a Branch of 2 elements *the xml branch is the group here_)
    * Element

here if we have a loop with more than 2 elements, we can't just increase the order and add the loop element to the group (Branch). 
we have to recalculate it to get the right index for the element.

PR #993, opened 2017-02-21
TDI-38212:tWebServiceInput returning empty result

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38212

### DIFF
--- a/main/plugins/org.talend.designer.components.libs/libs_src/tWebServiceInput/org/talend/DynamicInvoker.java
+++ b/main/plugins/org.talend.designer.components.libs/libs_src/tWebServiceInput/org/talend/DynamicInvoker.java
@@ -18,6 +18,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Vector;
+import java.util.Set;
 
 import javax.wsdl.Binding;
 import javax.wsdl.Operation;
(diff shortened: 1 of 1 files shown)


pyzhou-talend:
The element name(key) was different between response has multiple elements and single element. So it will get null if use the key of single element when it has multiple.

main/plugins/org.talend.designer.components.libs/libs_src/tWebServiceInput/org/talend/DynamicInvoker.java
```diff
@@ -337,16 +338,16 @@ public HashMap invokeMethod(String operationName, String portName, String[] args
         Map outputs = call.getOutputParams();
         HashMap map = new HashMap();
 
-        for (int pos = 0; pos < outNames.size(); ++pos) {
-            String name = (String) outNames.get(pos);
+            String name = (String) outNames.get(0);
```
jzhao-talend: Here `outNames` can't be empty ?

pyzhou-talend: It the name list of response elements, It can't be empty.

jzhao-talend: And seems the `outNames` didn't get as expect in customer's case:
```
[>checkVat>countryCode, >checkVat>vatNumber, >checkVatResponse>requestDate, >checkVatResponse>valid, >checkVatResponse>name, >checkVatResponse>address]
```

PR #996, opened 2017-02-22
feat(TDI-35388): handle ExpiredSecurityTokenException

chmyga:
* Add wrapper for OrganizationServiceStub

* Implement automatic token renewal

* Increment talend-mscrm lib version

### DIFF
--- a/main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/sdk/OrganizationServiceStubWrapper.java
+++ b/main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/sdk/OrganizationServiceStubWrapper.java
@@ -42,11 +42,18 @@ public class OrganizationServiceStubWrapper extends OrganizationServiceStub {
 
     private final String discoveryServiceUrl;
 
-    public OrganizationServiceStubWrapper(OrganizationServiceStub orgStub, MSCRMClient client, String discoveryServiceUrl)
+    private final int maxConnectionRetries;
+
+    private final int attemptsInterval;
+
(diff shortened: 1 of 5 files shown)


main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/sdk/OrganizationServiceStubWrapper.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private void renewToken() throws Exception {
+        orgStub = client.doGetOnlineConnection(discoveryServiceUrl);
+    }
+
+    private <T extends Exception, V extends XmlObject, R extends XmlObject> V perform(OperationProcessor<T, V, R> processor)
+            throws RemoteException, T {
+        V respDoc = null;
+        boolean connectionFixed = true;
+        int retries = 5;
```
jzhao-talend: After confirm with Pierre, This value should be customized on component side as  Marketo.
And need to add interval time between every retry also as marketo.
So please refer to what we have done in marketo.
```
         <PARAMETER
		NAME="MAX_RECONN_ATTEMPS"
		FIELD="TEXT"
		NUM_ROW="110"
		REQUIRED="true"
	>
		<DEFAULT>5</DEFAULT>
	</PARAMETER>

	<PARAMETER
		NAME="ATTEMPS_INTERVAL_TIME"
		FIELD="TEXT"
		NUM_ROW="110"
		REQUIRED="true"
		SHOW="false"
	>
		<DEFAULT>1000</DEFAULT>
	</PARAMETER>
```
`ATTEMPS_INTERVAL_TIME` is a hidden parameter

chmyga: Attempts interval time is not used at all. It is not shown in component. Do we really need to include it?

PR #1002, opened 2017-02-23
feat(TDI-35388): handle ExpiredSecurityTokenException

chmyga:
* Add reconnect timeout and retries parameters to MS CRM components

* Update mscrm library version in libraries pom.xml

* Update mscrm library version in components xml

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tMicrosoftCrmInput/tMicrosoftCrmInput_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tMicrosoftCrmInput/tMicrosoftCrmInput_begin.javajet
@@ -26,8 +26,14 @@ imports="
     String crmVersion = ElementParameterParser.getValue(node, "__MS_CRM_VERSION__");
 
     boolean isMsCrm2015OnPremise = "ON_PREMISE".equals(authenticationType) && "CRM_2015".equals(crmVersion);
+    
+    String attemptConnTimes = ElementParameterParser.getValue(node, "__MAX_RECONN_ATTEMPS__");
+	String attemptsInterval = ElementParameterParser.getValue(node, "__ATTEMPS_INTERVAL_TIME__");
 %>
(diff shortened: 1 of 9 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tMicrosoftCrmInput/tMicrosoftCrmInput_begin.javajet
```diff
@@ -26,8 +26,14 @@ imports="
...
 
     boolean isMsCrm2015OnPremise = "ON_PREMISE".equals(authenticationType) && "CRM_2015".equals(crmVersion);
+    
+    String attemptConnTimes = ElementParameterParser.getValue(node, "__MAX_RECONN_ATTEMPS__");
+	String attemptsInterval = ElementParameterParser.getValue(node, "__ATTEMPS_INTERVAL_TIME__");
 %>
 	int nb_line_<%=cid%> = 0;
+	
+	int reConnMaxNum_<%=cid%>=<%=attemptConnTimes%>;
+	reConnMaxNum_<%=cid%> = reConnMaxNum_<%=cid%> > 0 ? reConnMaxNum_<%=cid%> : 1;
```
jzhao-talend: Only one question about this one why can't be `0` ?
I think `0` means never retry connection when get error. That might be used by some user.
`client_<%=cid%>.setAttemptsInterval(<%=attemptConnTimes%>);`
I think this kind of set would ok. How do you think about this ?

chmyga: It's totally ok for me. I just made it in the same way as in marketo component. I don't mind changing it to something like that.

jzhao-talend: ah... yes.  Sorry I don't remember the reason for this. Maybe after discuss.

PR #1021, opened 2017-03-01
tdi 38212 upgrade the jar

pyzhou-talend:


pteyssier:
I prefer we don't have jar anymore in git and not compiled on developer machine.

PR #1035, opened 2017-03-03
fix(TDI-38351): fix for proxy

jzhao-talend:


main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/MSCRMClient.java
```diff
@@ -321,4 +326,30 @@ private static String discoverOrganizationUrl(DiscoveryServiceStub serviceStub,
...
+            proxyHost = System.getProperty("https.proxyHost");
+            proxyPort = System.getProperty("https.proxyPort");
+            proxyUser = System.getProperty("https.proxyUser");
+            proxyPwd = System.getProperty("https.proxyPassword");
+        }
+
+        if (proxyHost != null || proxyPort != null) {
+            proxyProps = new ProxyProperties();
+            proxyProps.setUserName((proxyUser == null ? "" : proxyUser));
+            proxyProps.setPassWord((proxyPwd == null ? "" : proxyPwd));
```
chmyga: password is not mandatory, so it can be set as null. But that's not critical, JFI.

main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/MSCRMClient.java
```diff
@@ -321,4 +326,30 @@ private static String discoverOrganizationUrl(DiscoveryServiceStub serviceStub,
...
+        HttpTransportProperties.ProxyProperties proxyProps = null;
+        // set by other components like tSetProxy
+        if (System.getProperty("https.proxyHost") != null) {
+            proxyHost = System.getProperty("https.proxyHost");
+            proxyPort = System.getProperty("https.proxyPort");
+            proxyUser = System.getProperty("https.proxyUser");
+            proxyPwd = System.getProperty("https.proxyPassword");
+        }
+
+        if (proxyHost != null || proxyPort != null) {
```
chmyga: Proxy port is also useless here without proxyHost value. so we can exclude it from condition.

main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/MSCRMClient.java
jzhao-talend: yes. you are right.

main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/MSCRMClient.java
jzhao-talend: agree with you

PR #1040, opened 2017-03-07
TUP-17232:Double Pipe delimiter - doesn't work

hcyi:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/model/process/jobsettings/JobSettingsManager.java
```diff
@@ -1219,7 +1224,7 @@ private static String findRealDbTypeForJDBC(IProcess process, String originalDbT
     }
 
     private static List<String> getMetadataChars() {
-        String[] metaChars = new String[] { "\\", "^", "$", ".", "?", "|", "[", "+", "*", "{", "(", ")", "}", "]" };
+        String[] metaChars = new String[] { "\\", "^", "$", ".", "?", "|", "[", "+", "*", "{", "(", ")", "}", "]", "||" };
```
nrousseau: not good at all...
we already have the character | in the list, and we know we must add some backslash for it.
What happens if one day one people add .. as separator, you'll add .. to the list???

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/model/process/jobsettings/JobSettingsManager.java
```diff
@@ -953,7 +953,12 @@ public static boolean isImplicittContextLoadActived(IProcess process) {
                 if (fileSparator.equals(File.separator)) {
                     fileSparator = TalendQuoteUtils.addQuotes("\\\\\\" + fileSparator); //$NON-NLS-1$
                 } else {
-                    fileSparator = TalendQuoteUtils.addQuotes("\\\\" + fileSparator); //$NON-NLS-1$
+                    char chs[] = fileSparator.toCharArray();
+                    String chBuffer = "";
+                    for (char ch : chs) {
+                        chBuffer = chBuffer + "\\\\" + ch; //$NON-NLS-1$
+                    }
+                    fileSparator = TalendQuoteUtils.addQuotes(chBuffer);
```
nrousseau: we need to add \ only if the character is one of the metadata bellow, why add for all???

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/model/process/jobsettings/JobSettingsManager.java
```diff
@@ -953,7 +953,12 @@ public static boolean isImplicittContextLoadActived(IProcess process) {
                 if (fileSparator.equals(File.separator)) {
```
nrousseau: maybe old code here, but why test with file separator?
I guess it was for windows for the \\ as separator, i don't think the / cause any issue for linux.
that means the code execution is different between OS

nrousseau:
This is really easy to test this with junits also, really can't merge this without junits

hcyi:
hi, check and redo , add some junits also . need your help try to review again,thx.

PR #1042, opened 2017-03-09
TUP-17214:Problem by using joblets from the reference project

hwang-talend:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/process/Process.java
```diff
@@ -3167,9 +3167,23 @@ public void removeUniqueConnectionName(String uniqueConnectionName) {
...
 
     public String generateUniqueNodeName(INode node) {
-        String baseName = node.getComponent().getName();
+        String baseName = getComponentName(node);
         return UniqueNodeNameGenerator.generateUniqueNodeName(baseName, uniqueNodeNameList);
     }
+    
+    private String getComponentName(INode node){
+        String baseName = node.getComponent().getName();
+        if (PluginChecker.isJobLetPluginLoaded()) {
```
nrousseau: If this is only for joblet component, why don't simply override the joblet component to change the name ???
This code make things even more complex / easy for bug, and dirty..

PR #1046, opened 2017-03-10
Feature/6.4/tup 16698 support context for hadoop conf jar

ycbai:


amarouni:
We had a quick look and it looks good, but can you add some short doc about the new feature like how it's supposed to work ? How can the user change the value of CONF_LIB in the tHadoopConfManager using context variables ?

nrousseau:
How can the user change the value of CONF_LIB in the tHadoopConfManager using context variables ? > only from UI actually, all the component part is totally hidden for the final user.

PR #1054, opened 2017-03-13
Bug/tbd 4639 t hive output sp to ss convert

lbourgeois:
When migrating from a Spark job item, getProcessFromItem first argument is a MapReduceProcessItemImpl so _if (ProcessItemImpl.class == item.getClass())_ test failed.

Need to upgrade it to if ((ProcessItemImpl.class == item.getClass()) || (ProcessItemImpl.class == item.getClass().getSuperclass())) for the Spark use case


nrousseau:
Hello,

Sorry we did the review a little late, but I reverted finally.

Issue is that normally we should call:
DesignerCoreService.getProcessFromItem(item)

we have 3 types of item:
(emf) ProcessItem ==> Linked to Process class
(emf) MapReduceProcessItem => linked to SparkProcess or MapReduceProcess
(emf) StormProcessItem => Linked to SparkStreamingProcess / StormProcess

the function above, will load the process class linked to the correct item.
But from the code mentioned above, you will always force to create the object Process (for DI jobs only), no matter the type of item who extends ProcessItem (even the routes from ESB extends the ProcessItem, so it makes things even more dangerous)

amarouni:
@nrousseau,

Sorry for the merge we had to do it as we had another tbd PR waiting for this one. @lbourgeois will take a look at the required changes.

Thanks,

PR #1073, opened 2017-03-16
TUP-17224:The Spark Property should not sync when deselect it in wizard

hcyi:


ycbai:
Please use HadoopClusterDragAndDropHandler to control whether use the spark properties from repository.

hcyi:
close this and push again .

PR #1083, opened 2017-03-20
fix(TDI-38351): set correct talend-mscrm version

dmytro-sylaiev:
* Fix wrong talend-mscrm lib version, it should be 2.1, not 2.0 for tos6.4


jzhao-talend:
The version is ok now. But I found that you didn't backport the modification of:
https://github.com/Talend/tdi-studio-se/commit/8828381c1349361909645c2246de4924e6923db6

iistomin-gl:
Please, could you fill the details in PR description to explain what you did.

jzhao-talend:
Because of the workflow. So maybe need buddy review also. @iistomin-gl

PR #1084, opened 2017-03-20
TUP-17240: Fix wrong Job's dependency list, when component supports multiple runtimes

ycbai:


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java
```diff
@@ -1111,74 +1106,69 @@ public boolean isDataAutoPropagated() {
     public List<ModuleNeeded> getModulesNeeded() {
         return getModulesNeeded(null);
     }
-    
+
 
     @Override
     public List<ModuleNeeded> getModulesNeeded(INode node) {
```
gonchar-ivan: FYI
Also as I saw from debug (maybe I'm wrong... it should be rechecked) getModulesNeeded() method is called during Studio launch for all TCOMP components. 
It had sense before, when components return the same RuntimeInfo (so it was initialized once during Studio launch)
But now it has no sense. Maybe consider to not call this method during Studio launch?

nrousseau: we need to check if we can change this indeed.. but in another task of course

PR #1092, opened 2017-03-22
TDI-38464: handle Error in code generation.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38464


pyzhou-talend:
The Try-Catch block of db2outbut added in TDI-35416 cause this issue. The try-catch is used to catch uncatched SQLException in whole db2output component, like Exception on creating table. But the exceptions external from branch does not have nested causes, so catch them is unnecessary.

PR #1105, opened 2017-03-29
Feature/6.4/tup 16749 provide guided tutorials

hcyi:


nrousseau:
need to review still 2 things:
- maybe we'll still need to download the videos instead of having them inside the studio (still question in review now, waiting answer)
- if we have the videos by default inside the studio, better don't commit those videos on git, but publish on nexus. (and videos will be downloaded during the build time)

PR #1108, opened 2017-03-29
Bugfix/6.4/tdi 38454 updating thashfile lib

dmytro-sylaiev:
Update thashfile library version after changes [in this PR](https://github.com/Talend/tdi-studio-se/pull/1088)

Set right architecture for this library (move files)

Update dependency in both component which use this library

Update pom.xml file in components.localprovider for adding this lib in next studio build.


main/plugins/org.talend.designer.components.libs/libs_src/thashfile/pom.xml
```diff
@@ -53,7 +53,7 @@
 	<build>
 		<resources>	
 			<resource>
-				<directory>./</directory>
+				<directory>src/main/java</directory>
```
gonchar-ivan: fix junit version to 4.12 please

PR #1113, opened 2017-03-30
bug(TDI-38421): Refix

dmytro-sylaiev:
* Make the checking for table existing the same as in the tTeradataOutput component

* Remove previous fix

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tCreateTable/tCreateTable_main.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tCreateTable/tCreateTable_main.javajet
@@ -2046,7 +2046,12 @@ if(columnList != null && columnList.size() > 0) {
 	            	whetherExist_<%=cid%> = true;
 	            }     
             <%
-            } else if("DB2".equalsIgnoreCase(dbType)){
+            } else if("DB2".equalsIgnoreCase(dbType) || "TERADATA".equalsIgnoreCase(dbType)){
+                    if ( "TERADATA".equalsIgnoreCase(dbType)) {
+            %>
+                String dbSchema_<%=cid%> = <%=ElementParameterParser.getValue(node, "__DBNAME__")%>;
+            <%
+                    }
             %>
                 java.sql.DatabaseMetaData dbMetaData_<%=cid%> = conn_<%=cid%>.getMetaData();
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tCreateTable/tCreateTable_main.javajet
```diff
@@ -2046,7 +2046,12 @@ if(columnList != null && columnList.size() > 0) {
 	            	whetherExist_<%=cid%> = true;
 	            }     
             <%
-            } else if("DB2".equalsIgnoreCase(dbType)){
+            } else if("DB2".equalsIgnoreCase(dbType) || "TERADATA".equalsIgnoreCase(dbType)){
+                    if ( "TERADATA".equalsIgnoreCase(dbType)) {
+            %>
+                String dbSchema_<%=cid%> = <%=ElementParameterParser.getValue(node, "__DBNAME__")%>;
```
dmytro-sylaiev: Teradata Manager doesn't have a dbSchema field unlike the DB2Manager one, also as managerfactory.. 
So if I don't want (and I don't want) to change the component's architecture and some API which can ruin some old jobs, I can do this way.

iistomin-gl: As I can see, 
```java
String dbSchema_<%=cid%>
```
local variable is declared only if dbType is TERADATA. But what happens if dbType is DB2 ? This local variable will not  be declared however it is used in code below. Will this code compile for DB2 ? I think it will not compile because the local variable dbSchema_* was not declared. Correct me if I'm wrong.

dmytro-sylaiev: As I mentioned in my comment, DB2's manager (inner class) has field dbSchema and global variable dbSchema_<%=cid%>, but TeradataManager doesn't have it, so due to I don't want to change the API of second one, I used this fix. 
For DB2 dbType I actually can not check (because of licence), but those variable was there for it before my changes.

iistomin-gl: Ok.

dmytro-sylaiev: Although, I can use the similar code in another else-if for only teradata, but it will be the code duplicate.

PR #1114, opened 2017-03-31
TDI-38466:Problem with writing time field with nulls to Exasol

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38466


pyzhou-talend:
It use method "setTimestamp" when the value is not null , so this modification should be OK. I've tested it on Exasol and Sybase.

PR #1129, opened 2017-04-07
feat(TDI-38091):add Microsoft CRM online 365/2016  odata client api

jzhao-talend:
fix refresh token doesn't work for read/delete and also improve code for insert/update


jzhao-talend:
**NO NEED REVIEW THIS ONE AT PRESENT**

jzhao-talend:
@undx Sorry bother you again.
Confirm with Pierre we move the source code to talend-mscrm. So need your review again.

PR #1133, opened 2017-04-10
TDI-38105:Update Excel components to support Excel 2016/next version

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38105


pyzhou-talend:
These 5 jars will be published to Nexus.

PR #1140, opened 2017-04-11
fix(TDI-38345): Implement migration task for NetSuite components

iistomin-gl:
* Refactored NewComponentMigrationTask to facilitate processing of element params
* Implemented migration task for NetSuite components

### DIFF
--- a/main/plugins/org.talend.designer.core.generic/plugin.xml
+++ b/main/plugins/org.talend.designer.core.generic/plugin.xml
@@ -48,6 +48,15 @@
                name="NewMarketoMigrationTask"
                version="6.4.0">
        </projecttask>
+       <projecttask
+               beforeLogon="false"
+               breaks="6.4.1"
+               class="org.talend.designer.core.generic.model.migration.NewNetsuiteMigrationTask"
+               description="Migration task for jobs using Netsuite components to TCOMP"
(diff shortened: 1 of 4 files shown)


main/plugins/org.talend.designer.core.generic/plugin.xml
```diff
@@ -1,207 +1,216 @@
...
-   <extension
-         point="org.talend.core.runtime.service">
-      <Service
-            class="org.talend.designer.core.generic.service.GenericService"
-            serviceId="IGenericService">
-      </Service>
-   </extension>
-
-</plugin>
+<?xml version="1.0" encoding="UTF-8"?>
```
dmytro-sylaiev: Something wrong with differences, and it make review quite harder.

iistomin-gl: Hmm ... I don't see such diff (deleting/re-adding) in Eclipse/IDEA IDE. An IDE shows normal diff which changed lines only. Probably, it is related to line ending (CR/LF), my git config uses autocrlf.

gonchar-ivan:
@nrousseau or @ycbai  
Could you please take a quick look too?
Following change was done: extract methods:
 processMappedElementParameter()
 processUnmappedElementParameter()
which allows to override them in inheritors.

PR #1143, opened 2017-04-12
fix(TDI-38564): tHttpRequest error message is not complete

wwang-talend:


main/plugins/org.talend.designer.components.localprovider/components/tHttpRequest/tHttpRequest_main.javajet
```diff
@@ -147,6 +145,27 @@ try{
...
+            int read_<%=cid%> = -1;
+        	try {
+            	while ((read_<%=cid%> = es_<%=cid%>.read(buf_<%=cid%>)) > 0) {
+            		baos_<%=cid%>.write(buf_<%=cid%>, 0, read_<%=cid%>);
+            	}
+        	} finally {
+        		es_<%=cid%>.close();
+        	}
+        
+        	responseMessage_<%=cid%> += ("\n" + new String(baos_<%=cid%>.toByteArray(), "UTF-8"));
```
pyzhou-talend: Why did you use the fixed charset "UTF-8" here? Is there any changes that some web server will return a response with other charset?  Will it cause a problem decode a stream of other charset with "UTF-8"?

wwang-talend:
strange, pengyu leave some comment here, but i can't see it. I will fix something by the comment

PR #1163, opened 2017-04-19
fix(TDI-38519): revert the order caching in the tFileOutputMSXML

akhabali:
This is for : 
https://jira.talendforge.org/browse/TDI-38519

We can't cache the xml element order because it's not a fix order and it's may change in branch
by caching the order we get bas insert position for the element causing IndexOutOfBoundsException.

this will revert the solution implemented here : 
https://jira.talendforge.org/browse/TDI-34433

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tFileOutputMSXML/tFileOutputMSXML_begin.inc.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tFileOutputMSXML/tFileOutputMSXML_begin.inc.javajet
@@ -370,16 +370,16 @@ class GenerateToolByDom4j{
 		org.dom4j.QName qname_<%=currEleName%>_<%=cid%> = org.dom4j.DocumentHelper.createQName("<%=localName%>",org.dom4j.DocumentHelper.createNamespace("<%=prefix%>","<%=uri%>"));
 		<%touchXMLNode.putCurrentElementByQName(currEleName);%>
 		
-		if(orderHelper_<%=cid%>.need_recalculate("<%=tool.connName + "_" + currEleName%>")) {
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tFileOutputMSXML/tFileOutputMSXML_begin.inc.javajet
```diff
@@ -391,48 +391,26 @@ class GenerateToolByDom4j{
...
-		if(orderHelper_<%=cid%>.tailAppend("<%=tool.connName + "_" + currEleName%>")) {
-			<%touchXMLNode.getXMLElement(parentName);%>.add(<%touchXMLNode.getXMLNode(currEleName);%>);
-		} else {
-			<%touchXMLNode.getXMLElement(parentName);%>.elements().add(orderHelper_<%=cid%>.index("<%=tool.connName + "_" + currEleName%>"),<%touchXMLNode.getXMLNode(currEleName);%>);
 		}
-		
-		orderHelper_<%=cid%>.setLastPorcessedField("<%=tool.connName + "_" + currEleName%>");
-		
+			
+		<%touchXMLNode.getXMLElement(parentName);%>.elements().add(order_<%=cid%>,<%touchXMLNode.getXMLNode(currEleName);%>);
```
gonchar-ivan: ';' inside <% %> seems strange for me. Could you please recheck is it correct?

akhabali: Yes it's working like that. (i didn't change this line)
it's seems like a java jet method call

PR #1178, opened 2017-04-24
TMDM-9123 : tMDMBulkload component usage causes high volume of select…

zfkang-talend:
… sqls to mdm db (add "insertOnly" parameter to control server side SELECT operation)


glzhou:
OK to merge

PR #1185, opened 2017-04-25
TUP-17557:All default setting are missing when deselecting "Use local

wchen-talend:
mode" in Spark Configuration


wchen-talend:
The same fix already done in https://github.com/Talend/tdi-studio-se/commit/123cee14febff75fcf733e4e16bcea1cedeb9dab

PR #1196, opened 2017-04-26
TDI-38545:Update version of "json-path" dependency for

pyzhou-talend:
"tExtractJSONFields" component to 2.1.0  (TESB-19182)

https://jira.talendforge.org/browse/TDI-38545


pyzhou-talend:
the build seems has passed. However there's first time I use this , not 100% sure about it.

PR #1198, opened 2017-04-26
fix(TDI-38441): set failOn default value to true tFileCopy

dmytro-sylaiev:
As simple as possible

Change default value of new property to 'true'


main/plugins/org.talend.designer.components.localprovider/components/tFileCopy/tFileCopy_java.xml
```diff
@@ -137,7 +137,7 @@
      FIELD="CHECK"
      NUM_ROW="70"
     >
-    <DEFAULT>false</DEFAULT>
+    <DEFAULT>true</DEFAULT>
```
pteyssier: this change the default behavior so it would need a migration task, this is not what was expected

dmytro-sylaiev:
Need to have a migration task instead of this changing.

PR #1202, opened 2017-04-26
fix(TDI-38625):update tMicrosoftCRM odata client api version

jzhao-talend:
For https://jira.talendforge.org/browse/TDI-38625


jzhao-talend:
Need merge after #1201 is merged

PR #1206, opened 2017-04-27
feat(tbd-4453): Add support for dynamic hadoop configuration - DO NOT MERGE 

rdubois:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/model/process/DataProcess.java
```diff
@@ -3043,57 +3043,52 @@ private void checkUseParallelize(INode graphicalNode) {
...
+        if (component != null) {
+            DataNode confNode = new DataNode(component, component.getName() + "_" + node.getUniqueName()); //$NON-NLS-1$
+            confNode.setActivate(node.isActivate());
+            confNode.setStart(true);
+            confNode.setSubProcessStart(true);
+            confNode.setDesignSubjobStartNode(confNode);
+            confNode.setProcess(node.getProcess());
+            if (ComponentCategory.CATEGORY_4_DI == componentCategory) {
+                IElementParameter clusterIdParam = confNode.getElementParameter("CLUSTER_ID"); //$NON-NLS-1$
+                clusterIdParam.setValue(hadoopClusterItemId);
```
ycbai: In fact "CLUSTER_ID" parameter was added to fix https://jira.talendforge.org/browse/TUP-17565. I think not only DI but also BD needs to it.

PR #1207, opened 2017-04-27
fix(TUP-17453): When convert a hadoop cluster to context, the context group is not added into MR/Spark job which uses it.

zwxue:


ycbai:
Please add junits for it.

PR #1210, opened 2017-04-27
Bug/6.4/tdi 38441 migration task

dmytro-sylaiev:


main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/AddFailOnErrorOnTFileCopyTDI38441.java
```diff
@@ -0,0 +1,81 @@
...
+    
+        
+	}
+    
+	
+
+	
+	private class AddFailOnErrorOptionConversion implements IComponentConversion {
+
+        private String field = "CHECK"; //$NON-NLS-1$
```
iistomin-gl: You can move these constants to enclosing class and declare them as normal constants (final and static).

But I think that it doesn't make sense to define constants for these string values because they are used only once.

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/AddFailOnErrorOnTFileCopyTDI38441.java
```diff
@@ -0,0 +1,81 @@
...
+            return ExecutionResult.FAILURE;
+        }
+    
+        
+	}
+    
+	
+
+	
+	private class AddFailOnErrorOptionConversion implements IComponentConversion {
```
iistomin-gl: I think that there is no need to define named inner class for this transformation code, this code can be placed in anonymous inner class:
```java
ModifyComponentsAction.searchAndModify(item, processType, filter, Arrays.<IComponentConversion>asList(new IComponentConversion() {

  @Override
  public void transform(NodeType node) {
    ...
  }
}));
```
Transformation logic is not complex and use of anonymous inner class makes reduces amount of code.

iistomin-gl:
Please, correct indentation.

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/AddFailOnErrorOnTFileCopyTDI38441.java
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                    ElementParameterType copyingDirectory = ComponentUtilities.getNodeProperty(node, isCopyDirectoryPropertyName);
+                    if (copyingDirectory != null) {
+                        if ("true".equals(copyingDirectory.getValue())) { //$NON-NLS-1$
+                            ComponentUtilities.setNodeValue(node, propertyName, "false"); //$NON-NLS-1$
+                            return;
+                        }
+                        ComponentUtilities.setNodeValue(node, propertyName, "true"); //$NON-NLS-1$
+                    }
+
```
lxia-talend: ComponentUtilities.setNodeValue(node, propertyName, "true");
should add it here, out of       if (copyingDirectory != null)

PR #1226, opened 2017-05-02
bugfix(TUP-17303)

Seloen:
Azure blob components should hide property type with use an existing
connection used
https://jira.talendforge.org/browse/TUP-17303


main/plugins/org.talend.repository.generic/src/main/java/org/talend/repository/generic/ui/DynamicComposite.java
```diff
@@ -425,4 +426,16 @@ public ConnectionItem getConnectionItem() {
     public void setConnectionItem(ConnectionItem connectionItem) {
         this.connectionItem = connectionItem;
     }
+
+    @Override
+    protected boolean isShouldDisParameter(IElementParameter curParam) {
```
nrousseau: this method is no use now, no? maybe we can just cancel all changes from this class

ycbai: no it is in use now.

PR #1238, opened 2017-05-04
fix(TUP-16778): keep the reference properties object when call Component.getCodegenPropInfos method

wwang-talend:


wwang-talend:
please don't merge it if it's not important as release is coming.

PR #1249, opened 2017-05-05
TMDM-9923 : Ability to control the timeout or start/end date of the transaction in MDM components

qcui:


zfkang-talend:
Validated.

PR #1277, opened 2017-05-15
fix(TDI-38773): tUniqRow's temporary file name pattern always the same.

wwang-talend:


pyzhou-talend:
And please change the commit name into "TDI-38773" when you do 'Squash and merge'.

PR #1279, opened 2017-05-15
TUP-17703:tMap - rename menu of Batch Change of Expressions

hcyi:


main/plugins/org.talend.designer.mapper/src/main/java/org/talend/designer/mapper/ui/visualmap/table/DataMapTableView.java
```diff
@@ -1015,9 +1015,9 @@ public void mouseDown(MouseEvent e) {
                 if (e.button == 3 && getZone() == Zone.OUTPUTS) {
                     Menu mainMenu = new Menu(tableForEntries);
                     tableForEntries.setMenu(mainMenu);
-                    // Custom
+                    // menu:apply routines
                     MenuItem customItem = new MenuItem(mainMenu, SWT.PUSH);
-                    customItem.setText("Custom");
+                    customItem.setText("Apply Routines");
```
nrousseau: better use i18n to allow translations

nrousseau: note that we apply only one routine, not several routines.. (even if applied to several columns)

hcyi: Change again , use the i18n rename to “Apply Routine”.

PR #1307, opened 2017-05-24
fix(TDI-38475) : change apiMode migration

undx:


undx:
As discussed with @pteyssier we remove apiMode to tMarketoConnection.
Closing PR.

PR #1319, opened 2017-05-30
fix(TDI-34726): Upgrade milyn-smooks framework.

iistomin-gl:
* The milyn-smooks framework upgraded to last version 1.7.0
* The tEDIFACTToXML input component modified to adapt to new milyn-smooks framework
* The required artifacts (milyn-smooks-all-1.7.0 and EDIFACT mappings) deployed to Nexus (https://talend-update.talend.com/nexus) and will be retrieved for the component

This PR depends on changes for tcommon-studio-se: https://github.com/Talend/tcommon-studio-se/pull/1048


dmytro-sylaiev:
As I know, it is a bad practice to add new .jar files into repository, it should be downloaded from nexus on client side

iistomin-gl:
The intent was to make minimal changes. Also, the https://wiki.talend.com/display/rd/Update+talend+component+libraries wiki page is about updating of Talend (not third party components). As I can see, it doesn't describe procedure for upgrading of 'libraries' plugins which provide third-party libs.

PR #1320, opened 2017-05-30
Update GeomFromText to support PostGIS > 1.5

yjacolin:
Starting to PostGIS 2.x, postgis function should be prefixed with ST_. These
functions are compatible with postgis 1.5.


fxprunayre:
This will avoid installing legacy.sql in Postgres database and will make life easier for users (https://www.talendforge.org/forum/viewtopic.php?id=54070). +1 for me.

pteyssier:
Hello, is there a jira related from talendforge explaining in details the context of this change?

yjacolin:
I don't think so as I am not aware of the specific Talend process for PR. Could you explain me how to help on this ticket? Thanks.

fxprunayre:
> Hello, is there a jira related from talendforge explaining in details the context of this change?

Yes. It would be good to have this added to coming release as we are relying on legacy Postgres functions now and it does not work anymore with latest PostgreSQL/Postgis version.
 
See https://jira.talendforge.org/browse/TDI-44129

@pteyssier can we expect a review of this ?

Fix for:
* https://github.com/talend-spatial/talend-spatial/issues/82
* https://jira.talendforge.org/browse/TDI-44129

See https://postgis.net/docs/reference.html

```
	
PostGIS has begun a transition from the existing naming convention to an SQL-MM-centric convention. As a result, most of the functions that you know and love have been renamed using the standard spatial type (ST) prefix. Previous functions are still available, though are not listed in this document where updated functions are equivalent. The non ST_ functions not listed in this documentation are deprecated and will be removed in a future release so STOP USING THEM.
```

PR #1322, opened 2017-05-31
fix(TDI-33454):Close the outputstream after write to the file.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-33454


main/plugins/org.talend.designer.components.libs/libs_src/talendExcel/src/main/java/org/talend/ExcelTool.java
```diff
@@ -293,6 +293,7 @@ public void setRowAccessWindowSize(int rowAccessWindowSize) {
 
 	public void writeExcel(OutputStream outputStream) throws Exception {
 		wb.write(outputStream);
```
wwang-talend: It's not a bug in my opinion.  And another thing, even if you want to close the stream, you need to consider the case when exception appear before call this method, so should call it at the finally block, you can use _finally.javajet.

PR #1327, opened 2017-05-31
TDI-38550 issue with nullable key attribute toracleoutput

mbasiuk-talend:
Override _generateSetBooleanForNullableKeyStmt_ Manager's method in OracleManager.
Jira issue:
https://jira.talendforge.org/browse/TDI-38550

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
+++ b/main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
@@ -2656,7 +2656,7 @@ public class CLASS {
             return true;
         }
         
-		protected String generateSetBooleanForNullableKeyStmt(String dynCol,Column column, int index, String incomingConnName, String cid,
+        protected String generateSetBooleanForNullableKeyStmt(String dynCol,Column column, int index, String incomingConnName, String cid,
                 int actionType) {
             String dynamicIndex=""+index;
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
```diff
@@ -2656,6 +2656,27 @@ public class CLASS {
...
+                prefix = "pstmt_";
+            } else if (actionType == INSERT_TYPE) {
+                prefix = "pstmtInsert_";
+            } else if (actionType == UPDATE_TYPE) {
+                prefix = "pstmtUpdate_";
+            }
+            StringBuilder setStmt = new StringBuilder();
+
+            setStmt.append(prefix + cid + ".setInt(" + dynamicIndex + ", ((" + incomingConnName + "." + column.getName()+"==null) "
+                    +"|| (" + incomingConnName + "." + column.getName()+".equals(\"\")))?1:0);\r\n");
```
wwang-talend: look strange to me for this code : it will be set to 1 when the input is null or empty string, and i think no need to care about empty string, so  no need to override this method.

and about this issue, i find we already have a property to support it, the name is [Support null in "SQL WHERE" statement], only need to check it, then it will be OK in my view.  And i check the history, maybe the reporter only want to ask us to make it checked as default(not sure we will do it).

PR #1331, opened 2017-06-01
fix(TDI-34993): tDB2Output Truncate Table FAILS

wwang-talend:


main/plugins/org.talend.designer.components.localprovider/components/templates/DB/Output/AbstractDBOutputFinally.javajet
```diff
@@ -31,6 +40,15 @@ imports="
 			if(resourceMap.get("conn_<%=cid%>")!=null){
 				try {
 					<%dbLog.conn().closeTry(null);%>
+					
+					<%
+					//need to end the transaction for db2 by the commit method before closing the connection, DDL and DML both need it
+					//maybe we need to rollback it when expcetion appear, but like all other db output components, we may commit something before it, so use commit here.
+					if(DB2.equalsIgnoreCase(dbmsId)){
+					%>
+                	((java.sql.Connection)resourceMap.get("conn_<%=cid%>")).commit();
```
pyzhou-talend: If we close the connection in the end part and this commit() method may cause Exception.

PR #1334, opened 2017-06-01
kjwang/fix(TUP-17701)Can't_setup_reopen_tMap_if_component_is_not_existing

kjwang-talend:
fix(TUP-17701)Can't_setup_reopen_tMap_if_component_is_not_existing
https://jira.talendforge.org/browse/TUP-17701

### DIFF
--- a/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
+++ b/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
@@ -2228,13 +2228,6 @@ public class Node extends Element implements IGraphicalNode {
             }
         }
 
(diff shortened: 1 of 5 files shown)


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/nodes/Node.java
```diff
@@ -2227,6 +2227,14 @@ public IMetadataTable getMetadataFromConnector(String connector) {
...
             }
         }
+
+        if (EConnectionType.FLOW_MAIN.equals(EConnectionType.getTypeFromName(connector))) {
+            for (IMetadataTable table : metadataList) {
+                if (table != null && EConnectionType.getTypeFromName(table.getAttachedConnector()) == null) {
+                    return table;
+                }
+            }
+        }
```
wchen-talend: After this modification it will always return the first table if can't find table by connector.
I think it's better to keep it as null if can't find it .

PR #1355, opened 2017-06-07
fix(TUP-17983): random issue when logon remote project error pop up

cmeng-talend:
fix(TUP-17983): random issue when logon remote project error pop up
https://jira.talendforge.org/browse/TUP-17983


cmeng-talend:
depend on https://github.com/Talend/tcommon-studio-se/pull/1068

PR #1369, opened 2017-06-09
fix(TBD-5163): Remove the dependencies signature in the fat jar

rdubois:


pteyssier:
It's not FAT, it just got big bones

PR #1376, opened 2017-06-12
TUP-17219:When set a spark job from repository to local mode and then uncheck the "use local mode" checkbox,the job returns to repository with a wrong state

hcyi:


hcyi:
Will re check this , and push again .

PR #1390, opened 2017-06-13
fix(TUP-16516): after change context group under repository re-import the whole project again will get error

cmeng-talend:
fix(TUP-16516): after change context group under repository re-import the whole project again will get error
https://jira.talendforge.org/browse/TUP-16516


cmeng-talend:
depend on https://github.com/Talend/tcommon-studio-se/pull/1088

PR #1403, opened 2017-06-15
fix(TUP-17996):When Deploying one customer's job in CI, the build failed with error: constant string too long

kjwang-talend:
fix(TUP-17996):When Deploying one customer's job in CI, the build failed with error: constant string too long
https://jira.talendforge.org/browse/TUP-17996

### DIFF
--- a/main/plugins/org.talend.designer.dbmap/src/main/java/org/talend/designer/dbmap/language/generation/DbGenerationManager.java
+++ b/main/plugins/org.talend.designer.dbmap/src/main/java/org/talend/designer/dbmap/language/generation/DbGenerationManager.java
@@ -70,6 +70,10 @@ public abstract class DbGenerationManager {
     protected String tabSpaceString = DEFAULT_TAB_SPACE_STRING;
 
     protected static final String DEFAULT_TAB_SPACE_STRING = ""; //$NON-NLS-1$
+    
+    protected List<String> queryColumnsSegments = new ArrayList<String>();
+    
+    protected List<String> querySegments = new ArrayList<String>();
 
     /**
      * DOC amaumont GenerationManager constructor comment.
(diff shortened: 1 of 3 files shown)


wwang-talend:
Look good except one place : 
We expect the query and query columns keep the same logic, it mean that we expect segments only, then we can only append the string in the javajet part, and no need to do more process, but for the query columns, it's not. For example : 
```
select id , name from test t
["select ", "id, ", "name " "from ", "test t"]
```
But for query column segments,
```
id , name
["id", "name"]
```
We have to append the "," in the javajet part, it may make some trouble when var exists.
so i think keep the same logic like query is better.

wwang-talend:
seem handleQueryEnd method is not useful, please consider the three cases : 
```
old mode :
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var1 + "aaaaaaaaaaaaaaaaaaaa" + context.var2 + "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var1 + "aaaaaaaaaaaaaaaaaaaa" + context.var2 + "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var3
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var1 + "aaaaaaaaaaaaaaaaaaaa" + context.var2 + "\"aaaa\".\"aaaa\" "

" + context.var3 + "

new mode of the current fix :
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var1 + "aaaaaaaaaaaaaaaaaaaa" + context.var2 + "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var1 + "aaaaaaaaaaaaaaaaaaaa" + context.var2 + "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var3
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa" + context.var1 + "aaaaaaaaaaaaaaaaaaaa" + context.var2 + "\"aaaa\".\"aaaa\" "

```
seems that the second and third one are not right.

wwang-talend:
after fixing the two things above, we can merge it

PR #1411, opened 2017-06-16
fix(TDI-37439): make the class name shorter

wwang-talend:


pyzhou-talend:
The second commit also looks fine.

PR #1436, opened 2017-06-22
fix(TDI-39037):Handle NPE for tReplaceList.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-39037


main/plugins/org.talend.designer.components.localprovider/components/tReplaceList/tReplaceList_begin.javajet
```diff
@@ -71,7 +71,9 @@ for (Object o : tMap_<%=lookup%>_<%=cid %>.keySet()) {
     
     <%=typeToGenerate_lookup%> search = tMap_<%=lookup%>_<%=cid %>.get(o).<%=lookupSearchName%>;
     <%=typeToGenerate_replacement%> replacement = tMap_<%=lookup%>_<%=cid %>.get(o).<%=lookupReplacementName%>;
-    replace_<%=cid%>.put(search,replacement);
+    if(search!=null&&replacement!=null&&!"".equals(search)){
```
jzhao-talend: Empty string value of `search` should be ignored?
I think `replacement` is null or not should both work.

pyzhou-talend: Yes, Empty string value of search should be ignored.
I will try to make it work when we have a vacant `replacement`

PR #1442, opened 2017-06-23
fix(TDI-32121):make "not null" fixed in db2 create table statement.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-32121


main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
```diff
@@ -451,7 +451,11 @@ public class CLASS {
                             createSQL.append(getAutoIncrement(column.getStartValue(), column.getStep()));
                         } else {
                             createSQL.append(setDefaultValue(column.getColumn().getDefault(), dataType));
-                            createSQL.append(setNullable(column.getColumn().isNullable()));
+                            if("ibmdb2_id".equalsIgnoreCase(getDBMSId())&&column.isKey()){
```
jzhao-talend: Better to use **DB2**  `public static final String DB2 = "ibmdb2_id";`

pyzhou-talend: ok, I will change it.

PR #1443, opened 2017-06-23
fix(TDI-33080): Implement ignoring of LogCatcher internal errors

iistomin-gl:
* Implemented functionality which allows to ignore internal error that can occur during processing of log messages


main/plugins/org.talend.designer.components.localprovider/components/tLogCatcher/tLogCatcher_begin.javajet
```diff
@@ -4,15 +4,20 @@
...
 
 <%
 	CodeGeneratorArgument codeGenArgument = (CodeGeneratorArgument) argument;
 	INode node = (INode)codeGenArgument.getArgument();
 	String cid = node.getUniqueName();
+	boolean ignoreInternalErrors = ("true").equals(ElementParameterParser.getValue(node.getProcess(), "__IGNORE_LOG_CATCHER_INTERNAL_ERRORS__"));
 
 %>
+	<%if(!ignoreInternalErrors){%>
+	try {
```
chmyga: It looks like checking the "Ignore LogCatcher internal errors" will lead to try-catch absence in the code. Is that supposed to work like that by design?

PR #1447, opened 2017-06-23
chore(TDI-37213): switch talend-codegen-utils lib to the latest 0.20.…

gonchar-ivan:
… version

Related to https://jira.talendforge.org/browse/TDI-37213

https://artifacts-oss.talend.com/nexus/content/repositories/TalendOpenSourceRelease/org/talend/libraries/talend-codegen-utils/0.20.0/ released

It has extended API (di enforcers were moved to it) in comparison with 0.17.0-SNAPSHOT
I am going to replace calls in codegen from Daikon to this library


gonchar-ivan:
@nrousseau 
could you please take a look or ask somebody to review? I'm not sure it is enough to change version only here.

PR #1450, opened 2017-06-26
igonchar/tdi 37213 call codegen-utils library in codegen instead of calling daikon

gonchar-ivan:
Related to https://jira.talendforge.org/browse/TDI-37213

Depends on https://github.com/Talend/tdi-studio-se/pull/1447


gonchar-ivan:
@wwang-talend , could you please take a look

PR #1487, opened 2017-07-05
fix(TDI-39054): improve the support for version and after the fix, no need to change the xml every release.

wwang-talend:


wwang-talend:
rebase to master for auto test

PR #1494, opened 2017-07-07
fix(TDI-39079) tFileInputExcel sheet position does not work for event model

jzhao-talend:
For https://jira.talendforge.org/browse/TDI-39079

1) Make sheet position work with event mode
2) Do position check when the position set not in the range of the file


pyzhou-talend:
And only suggestion, we can add simpleExcel.jar to the .gitignore file by changce.
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.libraries.excel/lib/.gitignore

PR #1495, opened 2017-07-07
fix(TUP-5360): Retrieve Salesforce Modules option is missing

ycbai:


main/plugins/org.talend.repository.generic/src/main/java/org/talend/repository/generic/action/GenericAction.java
```diff
@@ -53,15 +68,16 @@ protected void doRun() {
...
-            return;
-        }
         repositoryNode = getCurrentRepositoryNode();
         repObjType = (ERepositoryObjectType) repositoryNode.getProperties(EProperties.CONTENT_TYPE);
-        this.setText(getCreateLabel());
-        this.setToolTipText(getEditLabel());
-        Image nodeImage = getNodeImage();
+        if (compWizard == null) {
+            compWizard = GenericWizardServiceFactory.getGenericWizardInternalService().getComponentWizard(repObjType.getType(),
+                    null);
```
wwang-talend: the location is null, this will make a NPE when create a new connection metadata

PR #1503, opened 2017-07-10
Mbasiuk/tdi 31661 keep last thashoutput

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-31661


main/plugins/org.talend.designer.components.libs/libs_src/thashfile/src/main/java/org/talend/designer/components/hashfile/memory/AdvancedMemoryHashFile.java
```diff
@@ -1,94 +1,73 @@
...
+     * <li> MATCHING_MODE.KEEP_ALL: all records are stored</li>
+     * <li> MATCHING_MODE.KEEP_FIRST: if several records have the same key, then only first record is stored</li>
+     * <li> MATCHING_MODE.KEEP_LAST: if several records have the same key, then only last record is stored</li>
+     * </ul>
+     * 
+     * @param value to be stored in Cache.
+     * @return stored value or null if value shouldn't be placed in Cache.
+     */
+    public V put(V value) {
+        if (value != null) {
```
gonchar-ivan: I suggest to use here following refactoring
https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

We can distinguish exceptional and normal branches here.
Exceptional branch here is `value == null`
It is recommended to put all exceptional branches at the beginning of the method (guard clause, which guard method from exceptional argument/condition):
```
if (value == null) {
    return null;
}
```
Normal branches here are: 
1. matching mode is keep_all 
2. matching mode is keep_last or keep_first

Normal branches go after all guard clauses. We may use if-else for normal branches, if there are only 2 choices and they are both normal conditions. For 3 and more options better to use switch-case or several if statements.
I suggest to change this method in following way. Note, method clarity should be on the first place:
```
if (matchingMode == MATCHING_MODE.KEEP_ALL) {
    allList.add(value);
    return value;
}
if (matchingMode == MATCHING_MODE.KEEP_LAST) {
    firstOrLastHash.put(value, value);
    return value;
}
if (matchingMode == MATCHING_MODE.KEEP_FIRST && !firstOrLastHash.containsKey(value) ) {
    firstOrLastHash.put(value, value);
    return value;
}
// value wan't stored in cache
return null;
```

"Guard clauses" will remove nested conditions (nested conditions complicates code clarity).
Also I suggest to break keep_last and keep_first options into 2 branches despite code is the same. It is to highlight that these are different options.

gonchar-ivan: The last `return null` may go inside else statement. We return null in 2 cases:
1. when value is null
2. when value is not null, matching mode is keep_first and value was already stored.
So, if we put last return into else it should be more clear... What do you think?

main/plugins/org.talend.designer.components.libs/libs_src/thashfile/src/main/java/org/talend/designer/components/hashfile/memory/AdvancedMemoryHashFile.java
```diff
@@ -1,94 +1,73 @@
...
+                    firstOrLastHash.put(value, value);
+                    return value;
+                }
+            }
+        }
+        return null;
+    }
+
+    public Iterator<V> iterator() {
+        return matchingMode == MATCHING_MODE.KEEP_ALL ? allList.iterator() : firstOrLastHash.values().iterator();
```
gonchar-ivan: hm... I just found that this class is a candidate for splitting it into several subclasses:
KeepFirstHashFile, KeepLastHashFile, KeepAllHashFile

gonchar-ivan: However, this may require changing client code...
You may extract strategy instead:
AbstractStoringStrategy
KeepFirstStoring extends AbstractStoringStrategy
KeepLastStoring extends AbstractStoringStrategy
KeepAllStoring extends AbstractStoringStrategy

Then, you may do following in original class:
```
public class AdvancedMemoryHashFile<V> {
    
    private StoringStrategy storingStrategy;   
  
    public AdvancedMemoryHashFile(MATCHING_MODE matchingMode) {
         switch (matchingMode) {
         // create corresponding strategy
         }
    }

    public V put(V value) {
        return storingStrategy.put(value);
    }

    // etc
    
}
```

main/plugins/org.talend.designer.components.libs/libs_src/thashfile/src/main/java/org/talend/designer/components/hashfile/memory/AdvancedMemoryHashFile.java
```diff
@@ -1,94 +1,75 @@
...
+
+        if (matchingMode == MATCHING_MODE.KEEP_ALL) {
+            allList.add(value);
+            return value;
+        }
+        if (matchingMode == MATCHING_MODE.KEEP_LAST) {
+            firstOrLastHash.put(value, value);
+            return value;
+        }
+        if (!firstOrLastHash.containsKey(value)) firstOrLastHash.put(value, value);
```
gonchar-ivan: please, add curly braces.
http://www.oracle.com/technetwork/java/codeconventions-150003.pdf
section 7.4

main/plugins/org.talend.designer.components.libs/libs_src/thashfile/src/main/java/org/talend/designer/components/hashfile/memory/AdvancedMemoryHashFile.java
```diff
@@ -1,94 +1,75 @@
...
+        return value;
+    }
+
+    public Iterator<V> iterator() {
+        if (matchingMode == MATCHING_MODE.KEEP_ALL) return allList.iterator();
+        return firstOrLastHash.values().iterator();
```
gonchar-ivan: please add braces.

PR #1511, opened 2017-07-11
fixe(TUP-18118):Remote run in studio fails with file not found exception

wchen-talend:


main/plugins/org.talend.repository/src/main/java/org/talend/repository/ui/wizards/exportjob/handler/AbstractBuildJobHandler.java
```diff
@@ -188,10 +188,16 @@ protected String getProgramArgs() {
...
+        String property = System.getProperty("maven.addiitonal.params");
+        if (property != null) {
+            profileBuffer.append(SPACE);
+            profileBuffer.append(property);
+            profileBuffer.append(SPACE);
+        }
+        // profileBuffer.append(TalendMavenConstants.PREFIX_THREAD);
+        // profileBuffer.append(SPACE);
+        // profileBuffer.append(TalendMavenConstants.PREFIX_THREAD_VALUE);
+        // profileBuffer.append(SPACE);
```
nrousseau: please remove the unused comments here, no need to keep it

PR #1523, opened 2017-07-13
fix(TDI-39179): Fix handling of 'Ignore trailing zeros for BigDecimal'

iistomin-gl:
JIRA: https://jira.talendforge.org/browse/TDI-39179

* Fixed functionality related to handling of 'Ignore trailing zeros for BigDecimal' feature


iistomin-gl:
Actually, in KeyStruct the 'IGNORE_TRAILING_ZEROS_FOR_BIGDECIMAL' is also handled, please see https://github.com/Talend/tdi-studio-se/blob/e18a79a99f9963b26c997fa6d3206f951272b8ff/main/plugins/org.talend.designer.components.localprovider/components/tUniqRow/tUniqRow_begin.javajet#L95.

chmyga:
That is very confusing. Thanks for explanation!

PR #1527, opened 2017-07-14
fix(TDI-37182): Improve handling of parsing related errors.

iistomin-gl:
* Improved handling of value parsing related errors

JIRA: https://jira.talendforge.org/browse/TDI-37182

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tFileInputDelimited/tFileInputDelimited_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tFileInputDelimited/tFileInputDelimited_begin.javajet
@@ -291,7 +291,8 @@
     							}
     							%>
     							} catch(java.lang.Exception ex_<%=cid%>) {
-									rowstate_<%=cid%>.setException(ex_<%=cid%>);
+									rowstate_<%=cid%>.setException(new RuntimeException(String.format("Couldn't parse value for column '%s' in '%s'. %s",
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tFileInputDelimited/tFileInputDelimited_begin.javajet
```diff
@@ -812,7 +829,22 @@
     									<%
     									} else {
     									%>
-    										<%=targetConnName %>.<%=column.getLabel() %> = <%=defaultValue %>;
+											<%
+											if (!JavaTypesManager.NULL.equals(defaultValue)) {
```
chmyga: Correct me if I'm not right:
If we have a non-primitive java type, and the default value is "null"(String value) then we have to catch an exception which is thrown when we try to set it to the field. But why don't we need to catch the exception if it is not "null"? Do we really need this if-else block?

iistomin-gl: Actually, this line checks that default value is NOT "null". If default value is NOT "null" then we should catch exception (because devalue value is Java expression and can cause errors).

chmyga: Oh, now I see. thanks for the explaining! Everything is clear for me now.

PR #1543, opened 2017-07-20
fix(TDI-38195):make footer invisible and throw an exception for the

pyzhou-talend:
stream input
https://jira.talendforge.org/browse/TDI-38195

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tFileInputPositional/tFileInputPositional_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tFileInputPositional/tFileInputPositional_begin.javajet
@@ -776,7 +776,7 @@
             	}
             	%>
             	if(filename_<%=cid %> instanceof java.io.InputStream){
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tFileInputPositional/tFileInputPositional_begin.javajet
```diff
@@ -777,7 +776,7 @@
             	}
             	%>
             	if(filename_<%=cid %> instanceof java.io.InputStream){
-            		in_<%=cid %> = new java.io.BufferedReader(new java.io.InputStreamReader((java.io.InputStream)filename_<%=cid %>, <%=encoding %>));
+            		throw new java.lang.Exception("When the input source is a stream,footer shouldn't be bigger than 0.");
```
wwang-talend: RuntimeException is better than Exception in my view as you depend on the outside "try catch(ExceptionType )" now. Maybe it will not make error, but my advice is for safe.

PR #1544, opened 2017-07-20
fix(TDI-38703): can't run job successfully use new component

wwang-talend:


main/plugins/org.talend.libraries.custom/plugin.xml
```diff
@@ -6,7 +6,7 @@
               context="plugin:org.talend.libraries.custom"
               language="java"
               message="Library for components API"
-              name="talend-codegen-utils.jar" mvn_uri="mvn:org.talend.libraries/talend-codegen-utils/0.17.0-SNAPSHOT"
```
wwang-talend: can't change the name as this parts : 
https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/Component.java#L1188

https://github.com/Talend/tcommon-studio-se/blob/master/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/general/ModuleNeeded.java#L166

I am sure some bug exists here, but not about the current fix.

PR #1550, opened 2017-07-20
Dsylaiev/tdi 39089 fix uri encoding t file fetch

dmytro-sylaiev:
Hide 'Encode uri' setting when it is not needed


dmytro-sylaiev:
Yes, it is

PR #1551, opened 2017-07-21
fix(TDI-38635): tMicrosoftOutput can't be update null value with V2016 odata

jzhao-talend:
For https://jira.talendforge.org/browse/TDI-38635

This also depend on https://issues.apache.org/jira/browse/OLINGO-1114
which pull request have been created:
https://github.com/apache/olingo-odata4/pull/21


jzhao-talend:
Because of olingo still didn't give response about our pull request apache/olingo-odata4#21
We would patched the jar firstly.

PR #1555, opened 2017-07-24
fix(TUP-18099): assertConnectorTopologyCompatibility should not throw exception during design time

ycbai:


nrousseau:
I think it will be temporary only... this needs to be improved in a second step.
One of the issue is the check of all the use of jars from tcomp before design anything (like display jars from module view), but this is linked to another bug, so we can improve it again later.

PR #1575, opened 2017-08-01
feat(TUP-18319):Hide Wizard if TCOMP wizard is JDBC

hwang-talend:


ggu-talend:
I think, no merge for this commit only, after finished whole feature, will merge all. so close this one

PR #1583, opened 2017-08-03
fix TUP-18358:Drag oracle connection will change the query too

hcyi:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/process/TalendEditorDropTargetListener.java
```diff
@@ -1857,6 +1857,9 @@ private Command getPropertyPublicPart(RepositoryNode selectedNode, IElementParam
             IMetadataTable metadataTable = (IMetadataTable) selectedNode.getObject();
             command2.setTable(metadataTable);
         }
+        if (selectedNode.getProperties(EProperties.CONTENT_TYPE) == ERepositoryObjectType.METADATA_CONNECTIONS) {
```
ggu-talend: Maybe should be same as the line 1440?

hcyi:
Push again after review .

wchen-talend:
How about the change property type , it will override the sql also.

PR #1601, opened 2017-08-16
fix(TDI-39229): Fix lookup in tXMLMap when lookup source is document.

iistomin-gl:
* Fix lookup in tXMLMap when lookup source is document


wwang-talend:
close it as jira is closed already

PR #1607, opened 2017-08-23
fix(TDI-39289) : support tSnowflakeConnection in ELTJDBC

jzhao-talend:
For https://jira.talendforge.org/browse/TDI-39289

- Support tELTJDBCMap can use connection from tSnowflakeConnection component
- Make table action work in tELTJDBCOutput for snowflake.


nrousseau:
seems strange, the feature was more to be generic, and have the possibility to use any kind of jdbc component / tConnection into the eltmap.
And not have something specific only for snowflake.

I don't think it's the good solution here

jzhao-talend:
@nrousseau 
Do you means for the specific code for create table ?
About generate create table, because if we support "table action" on tELTJDBCOutput component.
We need to specify the mapping which I add in tcommon-studio-se/pull/1198
For snowflake there are some specify behavior for some code, So I add some special code for snowflake for generating create table SQL.

PR #1612, opened 2017-08-25
TMDM-9923 Ability to control the timeout or start/end date of the transaction in MDM components (remove dependency of routines.MDM)

lpw82gf:
Move MDM#getWSClientTimeout to TimeoutUtil.javajet to avoid depending on routines.MDM


pbabin:
LGTM

PR #1616, opened 2017-08-29
fix(TUP-18434) fix problem of refresh

nrousseau:
Fix the problem of reference display and limit the number of refresh
done.


nrousseau:
Do not merge now please, need to wait the linked commit to be merged as well, cf:
Talend/tcommon-studio-se#1207

PR #1624, opened 2017-09-04
fix (TUP-18492): Stats&Logs still use JTDS driver even select Microsoft

hcyi:


hcyi:
i will re create PR since the branch name contains specific characters.

PR #1632, opened 2017-09-13
feat(TUP-18531) Modify studio setup to only support java 8

nrousseau:


main/plugins/org.talend.repository/plugin.xml
```diff
@@ -2931,6 +2931,13 @@
               name="Configure the new INCLUDTRASHFILES property to 'true' in order to keep the same behavior as before"
               version="6.5.1">
         </projecttask>
+        <projecttask
+              beforeLogon="true"
+              class="org.talend.repository.model.migration.UpdateToJava8VersionMigrationTask"
```
ggu-talend: No break version to set?

PR #1645, opened 2017-09-26
feat(TUP-18311): Find a way to lower the cost of adding/upgrading BD Distros: have plugin for dynamic config with first implementation to read json / setup jars for jobs.

cmeng-talend:
https://jira.talendforge.org/browse/TUP-18311


cmeng-talend:
this pull requests depends on https://github.com/Talend/tcommon-studio-se/pull/1236

PR #1653, opened 2017-10-06
Ypiel/tdi39493 vertica copy stream

ypiel-talend:
**What is the current behavior? (You can also link to an open issue here)**
https://jira.talendforge.org/browse/TDI-39493

**What is the new behavior?**
Support of new options :
- COMPRESSION : BZIP | GZIP | LZO | UNCOMPRESSED
- TRAILING NULLCOLS
- ENCLOSED BY 'char'
- RECORD TERMINATOR 'string'
- SKIP nb records
- REJECTMAX
- ENFORCELENGTH
- ABORT ON ERROR
- STREAM NAME
- NO COMMIT

Moreover, the component now accept input flow with 'content' attribute than must be a java.io.FileInputStream like : [tFileInputRaw] ---> tVerticaBulkExec.

The virtual component tVerticaOuputBulkExec accept all new parameters except COMPRESSION and it can't deals with input streams.

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No


wwang-talend:
For almost all database bulk load components, we use the pattern above.

ypiel-talend:
The goal was to accept streams from a previous component as tFileInputRaw.content in tVerticaBulkExec since this a functionnality proposed by the Vertica driver : tFileInputRaw ---> tVerticaBulkExec
The tVerticaOutputBulkExec doesn't need to accept streams since, as other OBE components, it take rows and generate intermediate file.

wwang-talend:
i don't think we need this pattern(maybe i miss some information) : 
[tFileInputRaw] ---> tVerticaBulkExec, only need to do it like tfileinputdelimited for the stream mode.

ypiel-talend:
Hi Wang Wei,

The Jira you mention : https://jira.talendforge.org/browse/TDI-36640
has this comment :

> "the request here is to also support direct streaming without necessarily landing in a file [...] I would like to go directly to the bulk loader without going to the file system. [...]"

The component can now take a real  stream as input without the need of any file on disk. You can for example get the stream of a TGoogleDriveGet.content as input.

Do I miss something ?

wwang-talend:
I mean that from customer : 
support direct streaming without necessarily landing in a file

I think it's not necessary to support the mode :
input component can output stream as row value==>tverticabulkexec

I think the mode is enough for customer :
a text in tverticabulkexec which can set a stream object like this : 
```
(InputStream)(globalMap.get("a stream key"))
```
then only a single component : 
tverticabulkexec

your mode is better than my advice about supporting more than 1 stream, but there are some disadvantages : 
1. more complex than my advice, and not sure customer real need it
2. Talend only have several input components can work like tfileinputraw and TGoogleDriveGet, this is the limit, if customer want to pass other stream like from amazon s3, customer have to do this : 
use the tjavaflex to write code to create stream and also need to support the flow line
Of course, my advice is not easy too, and need to use tjava to create stream, and store it in globalMap, the mode is more easy for customer.

ypiel-talend:
Merge on local to resolve conflicts.

PR #1666, opened 2017-10-18
fix(TBD-5548): change default value for tOracleOutput varchar2 field

StefanLMNG:
The default length value for a VARCHAR2 field must be > 0.


StefanLMNG:
See comment on Jira : https://jira.talendforge.org/browse/TBD-5548

PR #1677, opened 2017-10-24
fix(TDI-39319): correct include/exclude types in REST API

undx:
https://jira.talendforge.org/browse/TDI-39319


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/migration/NewMarketoMigrationTask.java
```diff
@@ -110,6 +110,17 @@ protected ElementParameterType getParameterType(NodeType node, String paramName)
             if ("LEAD_SELECTOR".equals(paramName)) {
                 ParameterUtilTool.addParameterType(node, "CLOSED_LIST", "LEAD_SELECTOR_REST", String.valueOf(value));
             }
+            // Manage include/exclude REST types
+            if ("INCLUDE_TYPES".equals(paramName) || "EXCLUDE_TYPES".equals(paramName)) {
+                ElementParameterType whichApi = ParameterUtilTool.findParameterType(node, "USE_SOAP_API");
+                Object whichApiValue = ParameterUtilTool.convertParameterValue(whichApi);
```
akhabali: **wichApiValue** is ambiguous here . 
can you please rename that to `useSOAP `or `useREST `  regarding the test that you have bellow with the "true".equales(wichApiValue)

undx: sure.

PR #1681, opened 2017-10-26
fix(TUP-18738):UHG - Configuring SVN Polling from TAC

wchen-talend:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/preferences/PreferenceInitializer.java
```diff
@@ -161,9 +161,17 @@ public void initializeDefaultPreferences() {
...
+            boolean enableAutoCheckSvn = pAutoCheck == null ? true : Boolean.valueOf(pAutoCheck);
+            String pInterval = System.getProperty("svn.update.info.interval");
+            int svnCheckInterval = 1;
+            try {
+                svnCheckInterval = Integer.parseInt(pInterval);
+            } catch (NumberFormatException e) {
+                svnCheckInterval = 1;
+            }
+            store.setDefault(ITalendCorePrefConstants.SVN_UPDATE_INFO_AUTO_CHECK, enableAutoCheckSvn);
+            store.setDefault(ITalendCorePrefConstants.SVN_UPDATE_INFO_AUTO_CHECK_TIME_INTERVAL, svnCheckInterval);
```
nrousseau: i prefer force the value actually if set than set the default value only.
If do like this, after a patch, user will still need to modify manually again for every existing workspace.

ggu-talend: Yes, it's true, I think, for default value of preference, should be no condition. must be clear value.
I prefer set special value in loginTask or such

PR #1696, opened 2017-11-02
fix(TDI-39612)Talend 6.4 tExtractJSONFields not working similar to 6.2.1.

jzhao-talend:
For https://jira.talendforge.org/browse/TDI-39612

Add back old jsonpath lib to let custom chose which version do they prefer to use
`1.2.0` or `2.1.0`


main/plugins/org.talend.libraries.json/lib/.gitignore
```diff
@@ -0,0 +1,2 @@
+/json-path-1.2.0.jar
+/json-path-2.1.0.jar
```
pyzhou-talend: And I don't think we need this .gitignore, since this two jar will not involve in code compile.

jzhao-talend:
yes. @pyzhou-talend  you are right.

main/plugins/org.talend.libraries.json/lib/.gitignore
jzhao-talend: This is for local product test build,It doesn't strict., if you don't add those jars here for test.
This is not for product build for develop and test.

PR #1698, opened 2017-11-03
fix(TUP-18545): adapter the new interface with new parameter RuntimeContext which pass var from platform to tcomp

wwang-talend:


wwang-talend:
this should not affect the old trigger method as ivan make them both can work

PR #1704, opened 2017-11-07
fix(TDI-39653): tSocketOutput does not work with the ""

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-39653


main/plugins/org.talend.designer.components.localprovider/components/tSocketInput/tSocketInput_java.xml
```diff
@@ -112,7 +112,7 @@
       NUM_ROW="35"
       REPOSITORY_VALUE="ESCAPE_CHAR"
     >
-      <DEFAULT>"""</DEFAULT>
+      <DEFAULT>"\""</DEFAULT>
```
jzhao-talend: Seems in code generate '"' == '\"'

PR #1709, opened 2017-11-09
fix(TDI-39666): tFileInputExcel not processing Excel file -> Exception casued " java.io.IOException: Zip bomb detected!"

wwang-talend:


pteyssier:
Won't do. It would be a security issue.

PR #1715, opened 2017-11-10
fix(TDI-39682): support the multi line when code generation

wwang-talend:


wwang-talend:
close it as we have another way to fix it

PR #1724, opened 2017-11-20
fix(TDI-39190):Sheet name is not case sensitive.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-39190


pyzhou-talend:
OK. I will check it before merge. @wwang-talend

pyzhou-talend:
Excel doesn't have a Linux version. :) I will merge it.

PR #1725, opened 2017-11-20
fix(TDI-39707): mutli thread failed for stats and logs in redshift database by redshift jdb and postgresql output component

wwang-talend:


wwang-talend:
only fix it like we do for mysql, mssql and oracle

main/plugins/org.talend.designer.components.localprovider/components/tPostgresqlOutput/tPostgresqlOutput_main.javajet
```diff
@@ -164,9 +164,24 @@ skeleton="../templates/db_output_bulk.skeleton"
 <%
 			}
 		}
-	    %>
```
wwang-talend: not sure this part is useful, we only follow the same old logic for mysql, oracle,mssql, so maybe remove this part later.

wwang-talend: stats, log or meter support dynamic?

PR #1743, opened 2017-11-28
feat(TDI-39469): Implement marklogic migration task

dmytro-sylaiev:


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/migration/NewMarkLogicMigrationTask.java
```diff
@@ -0,0 +1,150 @@
...
+        List<NodeType> allNodes = getProcessType(item).getNode();
+        for (NodeType node : allNodes) {
+            if (sourceNodeUniqueName.equals(ComponentUtilities.getNodeUniqueName(node))) {
+                return node;
+            }
+        }
+        return null;
+    }
+
+    private static boolean doesConnectorsContains(List<ConnectionType> connectors, String metadataConnectorName) {
```
mbasiuk-talend: 1. Why this one static?
2. name is not good. e.g. isConnectorPresent(...).

main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/migration/NewMarkLogicMigrationTask.java
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+    }
+
+    private List<ConnectionType> getAllInputMainConnectors(NodeType nodeType) {
+        List<ConnectionType> allInputConnectors = ComponentUtilities.getNodeInputConnections(nodeType);
+        Iterator<ConnectionType> connectionIter = allInputConnectors.iterator();
+        while (connectionIter.hasNext()) {
+            ConnectionType nextConnector = connectionIter.next();
+            if (!nextConnector.getConnectorName().equals("FLOW")) {
```
mbasiuk-talend: Use `"FLOW".equals(...)`

main/plugins/org.talend.designer.core.generic/plugin.xml
```diff
@@ -57,6 +57,15 @@
                name="NewNetsuiteMigrationTask"
                version="6.4.0">
        </projecttask>
+       <projecttask
+               beforeLogon="false"
+               breaks="6.5.0"
```
mbasiuk-talend: new MarkLogic won't be in 6.5.1, do we really need this one to be present? What about 7.0.1 version?

PR #1744, opened 2017-11-29
fix(TDI-39743): Issue with tFileUnarchive with archives containing very large files

wwang-talend:


wwang-talend:
yes, we should remove the ant.jar if it's not useful, but need to do a check if it is used by other plugins in tup or other product like ESB, so will create a workitem for it later.

wwang-talend:
about use the maven origin path, i will try it.

pteyssier:
Correct remarks from Pengyu. But do not remove yet. The clean of duplicates is secondary and will be progressive.
Note: you sent me wrong pom with org.talend.libraries still inside.

main/plugins/org.talend.designer.components.localprovider/components/tFileUnarchive/tFileUnarchive_java.xml
```diff
@@ -87,7 +87,7 @@
    <CODEGENERATION>
      <IMPORTS>
        <IMPORT NAME="COMMONS-COMPRESS-1.10" MODULE="commons-compress-1.10.jar" MVN="mvn:org.talend.libraries/commons-compress-1.10/6.1.0"  UrlPath="platform:/plugin/org.talend.libraries.apache.common/lib/commons-compress-1.10.jar" REQUIRED="true" />
-       <IMPORT NAME="Ant-Zip" MODULE="ant.jar" MVN="mvn:org.talend.libraries/ant/6.0.0"  UrlPath="platform:/plugin/org.talend.libraries.apache/lib/ant.jar" REQUIRED="true" />
+       <IMPORT NAME="Ant-Zip" MODULE="ant-1.10.1.jar" MVN="mvn:org.apache.ant/ant/1.10.1" REQUIRED="true" />
```
pteyssier: the module should still be ant.jar.

PR #1745, opened 2017-11-29
fix(TUP-18492): Stats Logs still use JTDS driver even select Microsoft.

hcyi:


main/plugins/org.talend.repository/plugin.xml
```diff
@@ -2938,6 +2938,14 @@
               name="UpdateToJava8VersionMigrationTask"
               version="6.5.1">
         </projecttask>
+        <projecttask
+              beforeLogon="false"
+              breaks="6.3.2"
```
nrousseau: 6.3.0 (because we need a patch on 6.3.1)

main/plugins/org.talend.repository/plugin.xml
```diff
@@ -2938,6 +2938,14 @@
...
+              class="org.talend.repository.model.migration.ChangeJobsettingDBversionForMssqlMigrationTask"
+              id="org.talend.repository.model.migration.ChangeJobsettingDBversionForMssqlMigrationTask"
+              name="Fill the default db version of mssql for Jobsetting"
+              version="6.5.1">
```
nrousseau: 6.3.1 maybe? since we need patch on 6.3.1...  easy for backports, at least it'll be the same on each branches

PR #1754, opened 2017-12-04
TUP-19013 Studio hangs/freezes after clicking cancel while job is being

kjwang-talend:
TUP-19013 Studio hangs/freezes after clicking cancel while job is being built
https://jira.talendforge.org/browse/TUP-19013


main/plugins/org.talend.repository/src/main/java/org/talend/repository/ui/wizards/exportjob/scriptsmanager/BuildJobManager.java
```diff
@@ -274,7 +275,11 @@ public void run(IProgressMonitor wrMonitor) throws CoreException {
                 FilesUtils.copyFile(jobZipFile, jobFileTarget);
                 TimeMeasure.step(timeMeasureId, "Copy packaged file to target");
             } else {
-                throw new Exception(MAVEN_ERROR_MSG);
+                if (pMonitor != null && pMonitor.isCanceled()) {
```
ggu-talend: If want to enable to cancel the monitor, should be not only here, maybe need add for all monitor in each methods, like generateItemFiles, generateJobFiles, build, etc.

kjwang-talend: Modified it. Thanks

PR #1759, opened 2017-12-05
fix(TBD-5136): Support spaces in classpath for .sh jobs 

scoepeau:

- [ ] The commit(s) message(s) follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md#commit-message-format) ?
- [ ] Unit tests for the Java changes have been added (for bug fixes / features) ?
- [ ] TUJ for the JavaJet changes have been added (for bug fixes / features) ?
- [x] The new code does not introduce new technical issues

Space in classpath are not supported for the bash executable

**What is the new behavior?**

Spaces in the classpath are supported


kcoepeau:
@nrousseau I tested it in the Studio with a jobserver (simple case without spaces on its side) and with the client having spaces in its workspace path. 
I am not sure to understand what you mean by "good target OS".
Otherwise this change only affects when the user "Export" a job and for the .sh file only.

nrousseau:
as long as it's tested / working with a jobserver, i'm not against. But need to take care still since the lasts changes about quotes were made really few time ago here. cf issue: TPSVC-3862 (not even merged to 6.5 / master)

PR #1762, opened 2017-12-07
fix(TDI-37800):tWriteJsonField does not handle Arrays

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37800


pyzhou-talend:
This PR doesn't solve the issue.

PR #1770, opened 2017-12-11
fix(TDI-39503): tLDAPConnection error ( [LDAP: error code 13 - TLS confidentiality required]) when slapd is configured with require tls

wwang-talend:


wwang-talend:
need to update other ldap components too.

PR #1824, opened 2017-12-28
feat(TDI-39829):AWS S3 component support server-side encryption using KMS

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-39829


main/plugins/org.talend.designer.components.localprovider/components/tS3Put/tS3Put_messages.properties
```diff
@@ -6,7 +6,7 @@ ACCESS_KEY.NAME=Access Key
 SECRET_KEY.NAME=Secret Key
 BUCKET.NAME=Bucket
 KEY.NAME=Key
-FILE.NAME=File
+FILE.NAME=File or Stream
```
jzhao-talend: If change the the properties value, don't forget update babili:
https://talendforge.org/babili/index.php

PR #1838, opened 2018-01-04
feat(TUP-19046): add signature on items

ggu-talend:


main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/ItemSaveTriggerMigrationTask.java
```diff
@@ -0,0 +1,66 @@
...
+
+    //
+    // public List<ERepositoryObjectType> getTypes() {
+    // // all items
+    // return Arrays.asList((ERepositoryObjectType[]) ERepositoryObjectType.values());
+    // }
+
+    @Override
+    public Date getOrder() {
+        GregorianCalendar gc = new GregorianCalendar(2017, 12, 28, 12, 0, 0);
```
coheigea: Why is this date hard-coded to a particular value?

main/plugins/org.talend.repository/src/main/java/org/talend/repository/ui/actions/importproject/ImportDemoProjectItemsPage.java
```diff
@@ -347,7 +346,7 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException, Inte
                 } catch (Exception e) {
                     ExceptionHandler.process(e);
                 } finally {
-                    EmfResourcesFactoryReader.INSTANCE.getSaveOptionsProviders().remove(optionName);
+                    EmfResourcesFactoryReader.INSTANCE.removOption(demoImportOption);
                 }
```
coheigea: "removOption" should be "removeOption" if it's possible to change it without breaking other code.

PR #1855, opened 2018-01-11
fix(TDI-39395): tNetezzaNZload. Fix actualLogFilePath variable for context or global values

Voievidko:
https://jira.talendforge.org/browse/TDI-39395

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tNetezzaNzLoad/tNetezzaNzLoad_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tNetezzaNzLoad/tNetezzaNzLoad_begin.javajet
@@ -442,22 +442,17 @@ skeleton="../templates/db_output_bulk.skeleton"
 					}
 %>
 				}
-<%
-				String actualLogFilePath;
-				if (isLogFilePathSpecified) {
-					actualLogFilePath = parametersList.get(parametersList.indexOf("\"-lf\"") + 1);
-				} else {
-					actualLogFilePath = new StringBuilder("\"")
-										.append(dbtable.toUpperCase().replaceAll("\"", ""))
-										.append(".")
-										.append(dbuser.toUpperCase().replaceAll("\"", ""))
-										.append(".")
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tNetezzaNzLoad/tNetezzaNzLoad_begin.javajet
```diff
@@ -442,22 +442,17 @@ skeleton="../templates/db_output_bulk.skeleton"
...
-										.append(".")
-										.append(dbuser.toUpperCase().replaceAll("\"", ""))
-										.append(".")
-										.append(dbname.toUpperCase().replaceAll("\"", ""))
-										.append(".nzlog\"").toString();
-				}
-%>
-				nb_line_<%=cid%>.set(getProcessedRowsCount(<%=actualLogFilePath%>));
 
+				String actualLogFilePath = <%if(isLogFilePathSpecified){%>
```
wwang-talend: should append cid to avoid duplicated var name like this : 
actualLogFilePath_<%=cid%>

main/plugins/org.talend.designer.components.localprovider/components/tNetezzaNzLoad/tNetezzaNzLoad_begin.javajet
```diff
@@ -442,22 +442,17 @@ skeleton="../templates/db_output_bulk.skeleton"
...
+					<%=parametersList.get(parametersList.indexOf("\"-lf\"") + 1)%>;
+				<%} else {%>
+					"\"" + <%=dbtable%>.toUpperCase().replaceAll("\"", "")
```
wwang-talend: in my view, no useful now for replaceAll("\"", "") as we put the code to generated code instead of javajet code, for example : 
"value1","value2" is the value in javajet, so if we want to append them in javajet to "value1.value2", we need to replace the ", but now we put it to generated code, it's enough that : 
"\"" + "value1".toUpperCase() + "." + "value2".toUpperCase()

Another thing is best to use StringBuilder to avoid to create two much strings.

PR #1870, opened 2018-01-17
fix(TUP-19097)execute queries failed when change to context mode

xlwang-tlnd:
fix(TUP-19097)execute queries failed when change to context mode
https://jira.talendforge.org/browse/TUP-19097


main/plugins/org.talend.sqlbuilder/src/main/java/org/talend/sqlbuilder/erdiagram/ui/ErDiagramComposite.java
```diff
@@ -436,6 +441,8 @@ private String getMssqlCatalog(String realTableName, Connection conn) {
      * @return
      */
     private String getSelectStatement(List<String> tables, List<String> columns, List<String> wheres) {
+    
+    	String schema = this.getSchema();
```
hcyi: No need this code

PR #1900, opened 2018-01-24
fix(TDI-39827): some jars can't be imported or download after cxf upgrade

jzhao-talend:
For https://jira.talendforge.org/browse/TDI-39827

After upgrade CXF, this thow some wrong dependency of jar.
Provide download those jars.


pyzhou-talend:
After search, there's other components refer to these jars in tis provider. like xmlschema-core-2.2.1 in tMDMRestInput. Should we also change that?

PR #1902, opened 2018-01-24
feat(TBD-6262): Generate ps1 file when building a spark job

zafkirTalend:


main/plugins/org.talend.designer.runprocess/resources/templates/Job_run_template.ps1
```diff
@@ -0,0 +1,3 @@
+$fileDir = Split-Path -Parent $MyInvocation.MyCommand.Path
+cd $fileDir
+java ${talend.job.jvmargs.ps1} -cp ${talend.job.ps1.classpath} ${talend.job.class} ${talend.job.bat.addition} %*
```
kcoepeau: There is "talend.job.bat.addition"
Does it need to be changed to talend.job.ps1.addition ?

yu-talend: We can maybe do it later, because now I created only ${talend.job.jvmargs.ps1} ${talend.job.ps1.classpath} in maven pom. the others work well for this moment.

PR #1912, opened 2018-01-26
fix(TUP-19303):fix jdbc driver table repository update problem.

wchen-talend:


wchen-talend:
The rootProperties in the node , and properties in param are not the same object.
That is why we call getSubProperties().getProperties to set the value again .
Checked from the existing code . the simple param set value twice also .
       se.setStoredValue(value);
            if (value != null && !value.equals(oldValue)) {
                se = (Property<?>) getSubProperties().getProperty(se.getName());
                if (isDrivedByForm()) {
                    form.setValue(se.getName(), value);
                } else {
                    se.setStoredValue(value);
                }
                fireConnectionPropertyChangedEvent(newValue);
            }
I will check why there are not the same object.

PR #1913, opened 2018-01-26
fix(TDI-38023):compile error when setting applications with list.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38023


main/plugins/org.talend.designer.components.localprovider/components/tAmazonEMRManage/app_mapping.javajet
```diff
@@ -147,7 +147,7 @@ if(!use_customize_version){
 				%>
                 apps_<%=cid%>.add(new com.amazonaws.services.elasticmapreduce.model.Application()
                 <%if(parameterUtil.isValid(application.get("NAME"))){%>
-    			.withName(<%=application.get("NAME")%>)
+    			.withName("<%=application.get("NAME")%>")
```
jzhao-talend: This seems break value with double quote. Dose this parameter support double quote?

pyzhou-talend: This parameter is a list which doesn't have double quote and that lead to compile error. So I need to add the  double quote.

PR #1922, opened 2018-01-29
fix(TUP-19306):[bug] migration: all fields are writable for tJDBCInput

wchen-talend:
using repository metadata after migration


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/migration/NewComponentFrameworkMigrationTask.java
```diff
@@ -87,16 +87,26 @@ public void transform(NodeType nodeType) {
...
                 ComponentProperties compProperties = ComponentsUtils.getComponentProperties(newComponentName);
                 FakeNode fNode = new FakeNode(component);
+                ElementParameterType propertyParamType = getParameterType(nodeType, "PROPERTY:PROPERTY_TYPE");
+                boolean isRepository = false;
+                if (propertyParamType != null && "REPOSITORY".equals(propertyParamType.getValue())) {
+                    isRepository = true;
+                }
+                ElementParameterType sqlParamType = getParameterType(nodeType, "QUERYSTORE:QUERYSTORE_TYPE");
+                boolean isSQLRepository = false;
+                if (sqlParamType != null && "REPOSITORY".equals(propertyParamType.getValue())) {
```
wwang-talend: propertyParamType may null, so NPE appear

wwang-talend: should use sqlParamType ....

PR #2041, opened 2018-03-06
fix(TDI-40199): Migrate pagesize as number property instead of string

dmytro-sylaiev:

Job from 6.5 could become broken after migration because of string pagesizes
**What is the new behavior?**
Migration task force to change pagesize to number

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/model/migration/NewMarkLogicMigrationTask.java
```diff
@@ -66,35 +67,61 @@ protected void processUnmappedElementParameter(Properties props, NodeType nodeTy
...
+    private void setNumericPageSize(NodeType node, NamedThing property) {
+        String oldValue = ComponentUtilities.getNodeProperty(node, "PAGE_SIZE").getValue();
+        Property<Integer> pageSizeProperty = (Property<Integer>) property;
+        if (oldValue!= null && !oldValue.isEmpty()) {
+            oldValue = oldValue.replaceAll("\"", ""); //remove all double-quotes 
+            pageSizeProperty.setValue(Integer.valueOf(oldValue));
+        } else {
+            pageSizeProperty.setValue(DEFAULT_PAGE_SIZE);
+        }
+    }
```
olhava: Need to check if we can use Property valueEvaluator before parsing value to Integer. For the case of more complex expressions in old jobs.

olhava: And also use setStoredValue instead of setValue in case of expressions with globalMap and context.

PR #2114, opened 2018-03-20
fix(TCOMP-734): fix context and globalMap dynamic values wrapping in …

gonchar-ivan:
…quotes

**What is the new behavior?**

Add initial support for context and globalMap dynamic values

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:
https://jira.talendforge.org/browse/TCOMP-734


main/plugins/org.talend.sdk.component.studio-integration/jet_stub/generic/configuration.javajet
```diff
@@ -100,13 +101,16 @@ final java.util.Map<String, String> configuration_<%=cid%> = new java.util.HashM
                 <%=ElementParameterParser.getPasswordValue(node, p.getVariableName())%>
                 <%} else {
                     String value = ElementParameterParser.getStringElementParameterValue(p);
-                    value = org.talend.core.model.utils.TalendTextUtils.removeQuotes(value);
-                %>
-                <%=org.talend.core.model.utils.TalendTextUtils.addQuotes(value)%>
+                    if (new Utils().isDynamic(value)) {%>
```
chmyga: why do we create new Utils object for every variable? we could make that method static, or create one object for all configuration values.

main/plugins/org.talend.sdk.component.studio-integration/jet_stub/generic/utils.javajet
```diff
@@ -0,0 +1,34 @@
...
+class Utils {
+
+    /**
+     * Checks whether value contains dynamic expressions like context.var or globalMap.get("key")
+     */
+    boolean isDynamic(final String value) {
+        if (value == null) {
+            return false;
+        }
+        return ContextParameterUtils.isContainContextParam(value) || ContextParameterUtils.containCodeVariable(value, "globalMap.");
```
chmyga: Should we consider checking values for something like "input.column" or "row1.column"? This was possible in javajet components, AFAIK. It's totally fine for now. Just a thought for future.

chmyga:
Also, consider using the same approach for tables.

main/plugins/org.talend.sdk.component.studio-integration/jet_stub/generic/configuration.javajet
gonchar-ivan: We can't make a method static, because Utils class is a local class in method block.
But you are right I can change code to create a single object per component in the job.
Also I like an idea to move code to java.

main/plugins/org.talend.sdk.component.studio-integration/jet_stub/generic/utils.javajet
gonchar-ivan: You are right. I splitted the task. Adding support for connection name (record dynamic value) is not easy at the moment and probably not possible (not sure yet as I haven't finished analysis yet).
1. Connection name support also require possibility to reset configuration option per each record. Seems, it is not supported for the moment.
2. Another thing is I think we should collect all connection names available in the job. It is possible by invoking node.getProcess.getNodes(). For each node we may collect its connections. Value is dynamic, when it contains substring "someConnection.anything", where someConnection is any connection name present in the job.

gonchar-ivan:
1. We discussed possibility to move configuration code to java class. It is possible partially. We may extract following logic to separate class:
getParameterKey()
getParameterValue()

then in javajet we need to call
```
%>
configuration_<%=cid%>.put(<%=ref.getParameterKey(params)%>,<%=ref.getParameterValue(params)%>);
<%
```
but such class should be available for codegen plugin. Codegen plugin doesn't depend on any Tacokit specific code and initial plan was to not add any dependency on Tacokit code in Studio.

gonchar-ivan:
2. We decided to add "row1.column" value in separate task as it may be complex. As far as I know currently Tacokit runtime doesn't allow to reset configuration option value per each processed record.

PR #2117, opened 2018-03-21
TESB-21590:Pom generation for bundles seems generating the full code

csuntalendbj:


main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/maven/listener/ProcessChangeListener.java
```diff
@@ -117,7 +117,7 @@ private void casePropertiesChange(Object oldValue, Object newValue) {
                     } else if (!oldVersion.equals(property.getVersion())) {
                         // version change, will create new item
                         // create new job project.
-                        TalendJavaProjectManager.generatePom((ProcessItem) property.getItem());
+                        TalendJavaProjectManager.generatePom((ProcessItem) property.getItem(), 0x8);
```
nrousseau: use constants !!!!
how we can read this code later?

csuntalendbj: Changed， Added them in tesb-common-se repository

PR #2155, opened 2018-03-28
fix(TBD-6217) Fix hive application ID logging

yu-talend:
PR only to keep track on what has been done, to be finished.
All the components which called this file need to be re-checked, including:
tHiveInput, tHiveCreateTable (no statement actually executed, could be removed), tHiveLoad, tETLHiveOutput.


zlokomatic:
Hi,

why is this not merged and when will it be?
tHiveRow is unusable in Open Studio and this small fix is waiting since march!!

PR #2194, opened 2018-04-05
fix(TDI-36516): add quotes to table name

Voievidko:
https://jira.talendforge.org/browse/TDI-36516

**What is the new behavior?**
After adding quotes to table name, job executes as expected

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:


Voievidko:
Check this info in AWS docs:

**Delimited Identifiers**
Delimited identifiers (also known as quoted identifiers) begin and end with double quotation marks ("). If you use a delimited identifier, you must use the double quotation marks for every reference to that object. The identifier can contain any standard UTF-8 printable characters other than the double quote itself. Therefore, you can create column or table names that include otherwise illegal characters, such as spaces or the percent symbol.

**ASCII letters in delimited identifiers are case-insensitive and are folded to lowercase.** To use a double quote in a string, you must precede it with another double quote character.
https://docs.aws.amazon.com/redshift/latest/dg/r_names.html

PR #2317, opened 2018-05-16
fix(TUP-20127):[SVN] Job with oracle could not open after editing oracle connection.

zwxue:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/action/ReadProcess.java
```diff
@@ -61,7 +63,15 @@ protected void doRun() {
         Object obj = ((IStructuredSelection) selection).getFirstElement();
 
         RepositoryNode node = (RepositoryNode) obj;
-        ProcessItem processItem = (ProcessItem) node.getObject().getProperty().getItem();
+        Property property = node.getObject().getProperty();
```
nrousseau: Same as for the getLastVersion() in another PR.
If the RepositoryViewObject.getProperty() can return an invalid object, maybe need to fix it, or maybe it can be fixed directly by fixing the getLastVersion() or basic code in the repository

PR #2332, opened 2018-05-21
Igonchar/tcomp 816

gonchar-ivan:

Metadata migration is working during Studio start-up.

**What is the new behavior?**

Refactored TaCoKitMigrationModel and TaCoKitMigrationItemModel classes to move some actions (like migration, needsMigration etc) inside the class to make less chances to break class outside.
Implements Metadata migration during import (in applyMigrationTask overriden method)
Implements Process migration during import (as postImport action)
Reduce TaCoKitMigrationItemModel instances creation (before both TaCoKitMigrationItemModel and TaCoKitMigrationModel were created in many places. Now only TaCoKitMigrationModel is created)
Add TacokitComponentID parameter in Tacokit Component parameters (this ID is stored in process item). This is needed to simplify search for ComponentDetail during deserialization.

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:
https://jira.talendforge.org/browse/TCOMP-816


main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/metadata/handler/TaCoKitImportHandler.java
```diff
@@ -35,4 +51,67 @@ public TaCoKitImportHandler() {
...
+            getItem(importItem).ifPresent(connectionItem -> {
+                final TaCoKitConfigurationModel configuration = new TaCoKitConfigurationModel(connectionItem.getConnection());
+                if (configuration.needsMigration()) {
+                    try {
+                        manager().migrate(configuration, monitor);
+                        importItem.setProperty(connectionItem.getProperty());
+                        final IProxyRepositoryFactory factory = ProxyRepositoryFactory.getInstance();
+                        factory.save(connectionItem);
+                        final String version = RelationshipItemBuilder.LATEST_VERSION;
+                        RepositoryUpdateManager.updateDBConnection(connectionItem, version, false, false);
```
cmeng-talend: I remember that you said `RepositoryUpdateManager.updateDBConnection(connectionItem, version, false, false);` can't work during migration and we need to move it into `TaCoKitProcessMigrator#postImport`?

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/metadata/handler/TaCoKitProcessMigrator.java
```diff
@@ -0,0 +1,130 @@
...
+            ImportItem[] allImportItemRecords) {
+        // no-op
+    }
+
+    @Override
+    public void postImport(IProgressMonitor monitor, ResourcesManager resManager, ImportItem[] importedItemRecords) {
+        if (importedItemRecords == null) {
+            return;
+        }
+        for (final ImportItem importItem : importedItemRecords) {
```
cmeng-talend: I think maybe job migrations can be moved into TaCoKitImportHandler.java? and we only need to add `RepositoryUpdateManager.updateDBConnection(connectionItem, version, false, false);` here?

cmeng-talend: sorry, I think we need to move process migrations into TaCoKitMigrationManager

cmeng-talend: And maybe we also need to consider potential performance problem:
If we scan all items for tacokit migration when logging on project each time, it may be very slow, so I think maybe we need to save all tacokit component versions that studio loaded in somewhere, and check whether the versions changed or not when logging on project, if yes, we execute the tacokit migrations.

akhabali: IMHO, i think that there is no need to migrate the configuration at studio startup or user login.
for me the migration need to be done only when the serialized configuration is requested by a job or a wizard metadata. It's need to be lazy.
We can also make it user friendly, by prompting the user a popup informing him that the component will be migrated to the newer version.

gonchar-ivan: 1. Yes, it is possible that component version is upgraded without Studio version upgrade. This may happen, when we provide patch. Such patch may contain changes, which require migration.

2. Yes, I agree that migration should be lazy if possible and be done only when serialized configuration is requested. 
But I'm not sure it will be 100% requested for any possible operation in Studio.
E.g. user may build job without opening it. 
Another operation is pushing a job to TIC (I don't how it is done: .jar or configuration files).
Thus, I chose a safe approach.

gonchar-ivan: I need to consult with somebody to check whether lazy migration is ok for us

akhabali: For tacokit components. The migration is also executed at runtime. That's mean: if the job is executed with an older configuration and a newer version of the component. 
The migration will be performed at runtime (when component instance is created) and the job will run correctly.

At client side (studio, web), The migration is needed to update ui values to be aligned with the new version of the configuration and showed correctly in component properties form. which means that the migration is needed only when the configuration is requested by a ui.

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/metadata/handler/TaCoKitImportHandler.java
gonchar-ivan: yes, thanks. I will move it.

gonchar-ivan: Moved to separate ResourceHandler - TaCoKitRelatedItemsUpdater

PR #2461, opened 2018-06-28
TUP-20309 - integrate after group code part for tacokit

akhabali:
https://jira.talendforge.org/browse/TUP-20309

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/CodeGenerator.java
+++ b/main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/CodeGenerator.java
@@ -595,6 +595,13 @@ public class CodeGenerator implements ICodeGenerator {
                                     generateComponentCode(subProcess, node, ECodePart.PROCESS_DATA_END, incomingName, typeGen));
                         }
                     }
+
+                    if (isTacokitProcessor(node)) {
+                        codeComponent.append(generateComponentCode(subProcess, node, ECodePart.PROCESS_RECORDS_END, incomingName, typeGen));
+                        codeComponent.append(generateComponentCode(subProcess, node, ECodePart.PROCESS_DATA_BEGIN, incomingName, typeGen));
+                        codeComponent.append(generatesTreeCode(subProcess, node, ECodePart.MAIN, typeGen));
+                        codeComponent.append(generateComponentCode(subProcess, node, ECodePart.PROCESS_DATA_END, incomingName, typeGen));
+                    }
                     codeComponent.append(generatesTreeCode(subProcess, node, part, typeGen));
                     // if (isIterate) {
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/CodeGenerator.java
```diff
@@ -595,6 +595,13 @@ private StringBuffer generateComponentsCode(NodesSubTree subProcess, INode node,
                                     generateComponentCode(subProcess, node, ECodePart.PROCESS_DATA_END, incomingName, typeGen));
                         }
                     }
+
+                    if (isTacokitProcessor(node)) {
```
cmeng-talend: Hello @akhabali 
Seems that ECodePart.END_HEAD does the same thing;
Does ECodePart.END_HEAD can't meet your case?

cmeng-talend: It seems that the ECodePart.END_HEAD add codes at the same place with ECodePart.PROCESS_RECORDS_END

akhabali: No, this hook need to be inserted before the END and after the MAIN. and so before the END_HEAD

cmeng-talend: `ECodePart.END_HEAD` `ECodePart.END_BODY` `ECodePart.END_TAIL` are also inserted before the `ECodePart.END` and after the `ECodePart.MAIN`
From https://github.com/Talend/tdi-studio-se/blob/master/main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/CodeGenerator.java#L633-L641
As you can see, it is also add `generatesTreeCode(subProcess, node, ECodePart.MAIN, typeGen)`, Codes of `ECodePart.PROCESS_RECORDS_END` can be put into `ECodePart.END_HEAD`, but I am not sure whether it can work well for you, since it adds more codes like `generateComponentCode(subProcess, node, ECodePart.MAIN, incomingName, typeGen)`

akhabali: The code you point here don't match what i'm trying to achieve here as it do more things.
Also we already have those code part in our processor that i don't want to break.
I prefer to keep my isolated hook, to control exactly what is generated and when it's generated.

cmeng-talend: ok

PR #2468, opened 2018-07-02
fix(TDI-32185): fix limit for dynamic columns

chmyga:
* Add one line to limit

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:


pyzhou-talend:
I think maybe we need a migrationtask for this issue. Although it's wrong behavior in the past, we should always keep the old behavior for old job, right?

chmyga:
We don't change anything in component itself. We don't add any configuration parameters. We change only behavior in the task itself. So there's no need in migration task I think. Correct me if I'm wrong - we need a migration task only for components, which have new parameters in configuration part.

pyzhou-talend:
Maybe you are right. I agree with you deeply from my heart. However I'm not so certain about this. @jzhao-talend would know this well. Let's waite his comment and approve.

chmyga:
Added migration task, which decrements limit, to make old jobs work as they worked before.

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/DecrementLimitTFileInputDelimitedTDI32185.java
```diff
@@ -0,0 +1,77 @@
...
+        						ColumnType columnType = (ColumnType) columnObject;
+        						if(columnType.getType().equals("id_Dynamic")) {
+        							hasDynamic = true;
+        						}
+        					}
+        					if(!hasDynamic) {
+        						return;
+        					}
+        					ElementParameterType limit = ComponentUtilities.getNodeProperty(node, "LIMIT");
+        					Integer intLimitValue = Integer.parseInt(limit.getValue());
```
jzhao-talend: I think this migration can't work with empty/context/globalMap/... kinds of value.
Maybe you can add "-1" after the parameter value. like:
If customer fill `context.lilimit`, it can be migrate to  `context.lilimit-1`

jzhao-talend: Maybe need confirm with Pierre about this kind of solution.
This kind of solution maybe confuse customer.

chmyga:
Added a fix for this. We can discuss this with Pierre and I will revert it if it is not required.

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/DecrementLimitTFileInputDelimitedTDI32185.java
```diff
@@ -59,10 +60,17 @@ public void transform(NodeType node) {
...
-        						limit.setValue(String.valueOf(intLimitValue));
+        					try {
+        						Integer intLimitValue = Integer.parseInt(limit.getValue());
+        						if(intLimitValue > 0) {
+        							intLimitValue--;
+        							limit.setValue(String.valueOf(intLimitValue));
+        						}
+        					} catch(NumberFormatException e) {
+        						String sLimit = limit.getValue();
+        						if(ContextParameterUtils.isContainContextParam(sLimit)) {
```
jzhao-talend: why only consider context value?  no globalMap or other kinds of expression

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/DecrementLimitTFileInputDelimitedTDI32185.java
```diff
@@ -0,0 +1,81 @@
...
+import org.talend.designer.core.model.utils.emf.talendfile.ElementParameterType;
+import org.talend.designer.core.model.utils.emf.talendfile.MetadataType;
+import org.talend.designer.core.model.utils.emf.talendfile.NodeType;
+import org.talend.designer.core.model.utils.emf.talendfile.ProcessType;
+
+public class DecrementLimitTFileInputDelimitedTDI32185 extends AbstractJobMigrationTask {
+
+	@Override
+	public Date getOrder() {
+        GregorianCalendar gc = new GregorianCalendar(2018, 12, 10, 0, 0, 0);
```
jzhao-talend: i remember that month number should be `11`

jzhao-talend:
Others LGTM, see my last comment.

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/DecrementLimitTFileInputDelimitedTDI32185.java
chmyga: fixed in the latest commit.

PR #2548, opened 2018-07-27
TUP-20482 : active if with new legth strategy for comparing values.

akhabali:
This work still having issue with Tables

https://jira.talendforge.org/browse/TUP-20482

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/TaCoKitElementParameter.java
+++ b/main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/TaCoKitElementParameter.java
@@ -85,7 +85,7 @@ public class TaCoKitElementParameter extends ElementParameter implements IAdditi
         pcs.firePropertyChange(name, oldValue, newValue);
     }
 
-    public void fireValueChange(final Object oldValue, final Object newValue) {
+    void fireValueChange(final Object oldValue, final Object newValue) {
         for (final IValueChangedListener listener : valueChangeListeners) {
             listener.onValueChanged(this, oldValue, newValue);
         }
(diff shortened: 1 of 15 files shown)


main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/TaCoKitElementParameter.java
```diff
@@ -186,22 +186,22 @@ public void cloneAddionalInfoTo(final IAdditionalInfo targetAdditionalInfo) {
     public void setValue(final Object newValue) {
         final Object oldValue = super.getValue();
         super.setValue(newValue);
-        firePropertyChange(getName(), oldValue, newValue);
+        firePropertyChange("value", oldValue, newValue);
```
gonchar-ivan: changing to "value" will break Suggestions method. see ActionParametersUpdater
However, we can consider to change logic in Suggestions or
we can file 2 events:
1. with getName()
2. with "value"

gonchar-ivan: Aha, found a fix below. Thanks!

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/ui/composite/TaCoKitComposite.java
```diff
@@ -57,12 +54,11 @@
 
     private List<? extends IElementParameter> parameters;
 
-    private PropertyChangeListener paramChangedListener = new PropertyChangeListener() {
-
-        @Override
-        public void propertyChange(final PropertyChangeEvent evt) {
-            refresh();
+    private PropertyChangeListener redrawListener = evt -> {
+        if (!"show".equals(evt.getPropertyName())) {
```
gonchar-ivan: I would suggest to rename "show" -> "refresh".
However, it is not critical

akhabali: the property name is the name of the field that changed in the class that own the listeners and which is the field "show" here.

PR #2593, opened 2018-08-10
fix(TBD-7495): Add checkbox to transform date type in timestamp

kcoepeau:

- [ ] The commit(s) message(s) follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md#commit-message-format) ?
- [ ] Unit tests for the Java changes have been added (for bug fixes / features) ?
- [ ] TUJ for the JavaJet changes have been added (for bug fixes / features) ?
- [ ] The new code does not introduce new technical issues

Behavior was changed in Studio 6.3.2 to use Timestamp instead of Date

**What is the new behavior?**

Checkbox is added to be able to use old behavior and use Date type.

Checkbox is checked by default to use Timestamp type.


main/plugins/org.talend.repository/plugin.xml
```diff
@@ -3093,6 +3093,15 @@
               name="RenameELTJDBCComponentsMigrationTask"
               version="7.1.1">
         </projecttask>
+        <projecttask
+              beforeLogon="false"
+              breaks="6.3.2"
+              class="org.talend.repository.model.migration.spark.UseSqlDateTypeInDataframesTask"
+              description="Keep Date type for component impacted by change from Date to Timestamp"
+              id="org.talend.repository.model.migration.Aspark.UseSqlDateTypeInDataframesTask"
```
ostoliarov: Shouldn't the **id** here be the same as **class**?

PR #2601, opened 2018-08-13
TCOMP-991: ignore activations from config not being part of the form …

akhabali:
…while creating metadata

https://jira.talendforge.org/browse/TCOMP-991

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SettingVisitor.java
+++ b/main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SettingVisitor.java
@@ -33,8 +33,11 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.TreeMap;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.talend.core.model.process.EComponentCategory;
 import org.talend.core.model.process.EConnectionType;
 import org.talend.core.model.process.EParameterFieldType;
@@ -69,6 +72,8 @@ import org.talend.sdk.component.studio.util.TaCoKitUtil;
  * Creates properties from leafs
  */
 public class SettingVisitor implements PropertyVisitor {
+
+    private final static Logger LOGGER = LoggerFactory.getLogger(SettingVisitor.class.getName());
 
     /**
      * Specifies row number, on which schema properties (schema widget and guess schema button) should be displayed
@@ -507,12 +512,14 @@ public class SettingVisitor implements PropertyVisitor {
 
         node.getProperty().getCondition()
                 .forEach(c -> {
-                    final String targetPath = pathResolver.resolvePath(node.getProperty().getPath(), c.getTarget());
-                    if (targetPath != null && !targetPath.isEmpty()) {
-                        c.setTargetPath(targetPath);
+                    final Optional<String> targetPath = pathResolver.resolvePath(node.getProperty().getPath(), c.getTarget());
+                    if (targetPath.isPresent()) {
+                        c.setTargetPath(targetPath.get());
                         activations.computeIfAbsent(origin.getProperty().getPath(), (key) -> new HashMap<>());
                         activations.get(origin.getProperty().getPath()).computeIfAbsent(level, (k) -> new ArrayList<>());
                         activations.get(origin.getProperty().getPath()).get(level).add(c);
+                    } else {
+                        LOGGER.warn("can't find target property '" + c.getTarget() + "' used on property '" + c.getSourcePath() + "'");
                     }
                 });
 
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/resolver/AbsolutePathResolver.java
```diff
@@ -82,11 +79,10 @@ private String doResolvePath(final String nodePath, final String parameterRefere
             while (ref.startsWith("..")) {
                 final int lastDot = current.lastIndexOf('.');
                 if (lastDot < 0) {
-                    ref = "";
-                    break;
+                    return null; // property not found
```
gonchar-ivan: Probably it is worth to change return type to Optional<String> and return Optional.empty() here.
This will force to implement "null check" in client code.

gonchar-ivan: I copy-paste this code from component-form-core. It will be good to make corresponding change there to align with web.
(Ideally, we need to extract this class to some common module/lib

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SettingVisitor.java
```diff
@@ -507,10 +507,13 @@ private void buildActivationCondition(final PropertyNode node, final PropertyNod
...
-                    activations.computeIfAbsent(origin.getProperty().getPath(), (key) -> new HashMap<>());
-                    activations.get(origin.getProperty().getPath()).computeIfAbsent(level, (k) -> new ArrayList<>());
-                    activations.get(origin.getProperty().getPath()).get(level).add(c);
+                    final String targetPath = pathResolver.resolvePath(node.getProperty().getPath(), c.getTarget());
+                    if (targetPath != null && !targetPath.isEmpty()) {
+                        c.setTargetPath(targetPath);
+                        activations.computeIfAbsent(origin.getProperty().getPath(), (key) -> new HashMap<>());
+                        activations.get(origin.getProperty().getPath()).computeIfAbsent(level, (k) -> new ArrayList<>());
+                        activations.get(origin.getProperty().getPath()).get(level).add(c);
+                    }
```
gonchar-ivan: Not sure it is ok to ignore unresolved parameter silently.
At least it is worth to log it in Debug mode

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/resolver/AbstractParameterResolver.java
```diff
@@ -71,29 +70,31 @@ public void resolveParameters(final Map<String, IElementParameter> settings) {
...
-                    if (redrawParameter != null) {
-                        parameter.setRedrawParameter(redrawParameter);
-                    }
-                    final String callbackProperty = parameter.getName().replaceFirst(absolutePath, parameterRoot.getPath());
-                    final IActionParameter actionParameter = parameter.createActionParameter(callbackProperty);
-                    action.addParameter(actionParameter);
-                });
+                if (absolutePath != null && !absolutePath.isEmpty()) {
+                    final List<TaCoKitElementParameter> parameters = findParameters(absolutePath, settings);
+                    final SimplePropertyDefinition parameterRoot = expectedParameters.next();
```
gonchar-ivan: Potentially there is an error.
Consider following:
` parameters = {"incorrectTarget", "correct/target"}`
and
action method declaration:
`action(@Option("i") String incorrect, @Option("c") String correct)`

`incorrectTarget` may be ignored, however `i` parameter in action method is not ignored.
Thus, on 2nd iteration current code will try to map `correct/target` to `i` parameter.

(maybe need to call expectedParameters.next() in `else` branch)

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SettingVisitor.java
akhabali: I will add an error log here, but, i'm for a fail fast paradigm. 
Here, it's to late to catch this kind of errors. We may add some more validations to the component model (maven plugin and container creation)

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/resolver/AbstractParameterResolver.java
akhabali: Same thing here we can't allow bad parameters. This should never happen. We need to validate this at compile time or at deployment phase as a late verification.

**Note that this PR is for the ActiveIf** that can have target that comes from the parent configuration which will not be present in metadata wizard. as we can only create a part of the configuration.
like creating a datastore without having the dataset properties.

class MyDataStore{}

class MyDataSet {
 @Option
 boolean requireDatastore;

 @Option
 @ActiveIf(target = "requireDatastore", value ="true")
 MyDataStore datastore;
} 

This fails when you will try to show datastore wizard as it get the activation from the dataset.

akhabali:
outdated

PR #2687, opened 2018-09-11
TESB-23382 New "REST API Definitions" metadata node

dsergent:
**What is the new behavior?**
Added extension point to contribute OAS decoder(s).
Added new message to console for ESB jobs/routes execution.

- [ ] Bugfix
- [X] Feature

- [] Yes
- [X] No

**Other information**:


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/cmd/ChangeValuesFromRepository.java
```diff
@@ -326,6 +326,10 @@ public void execute() {
                 }
             }
             Map<Object, Object> contextMap = new HashMap<Object, Object>();
+            if (elem instanceof INode) {
+                contextMap.put("NODE", (INode) elem);
+            }
+
```
sbouchet: used to populate component table with metadata node infos. @nrousseau do you know another way ?

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/cmd/ChangeValuesFromRepository.java
```diff
@@ -410,6 +414,7 @@ else if (connection instanceof SalesforceSchemaConnection && "CUSTOM_MODULE_NAME
                     } else {
                         IMetadataTable metaTable = table;
                         if (metaTable == null && elem instanceof Node) {
+                            contextMap.put("NODE", elem);
```
sbouchet: same as above.

main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/RunProcessContext.java
```diff
@@ -1950,17 +1889,60 @@ public void loadLog4jContextFromProcess() {
...
+                        .setValue(RunProcessPrefsConstants.LOG4JLEVEL, (String) log4jLevelParam.getValue());
             } else {
                 RunProcessPlugin.getDefault().getPreferenceStore().setValue(RunProcessPrefsConstants.LOG4JLEVEL, "");
             }
         }
-        setLog4jLevel(RunProcessPlugin.getDefault().getPreferenceStore().getString(RunProcessPrefsConstants.LOG4JLEVEL));
+        setLog4jLevel(
+                RunProcessPlugin.getDefault().getPreferenceStore().getString(RunProcessPrefsConstants.LOG4JLEVEL));
+    }
+
```
sbouchet: below is a method that add a message in the console for ESB components only that display the endpoint url .

main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/RunProcessContext.java
```diff
@@ -637,14 +652,18 @@ public void run() {
...
+                                                String welcomeMsg =
+                                                        mf.format(new Object[] { process.getLabel(), new Date() });
                                                 processMessageManager
-                                                        .addMessage(new ProcessMessage(MsgType.CORE_OUT, welcomeMsg + "\r\n")); //$NON-NLS-1$
+                                                        .addMessage(new ProcessMessage(MsgType.CORE_OUT,
+                                                                welcomeMsg + "\r\n")); //$NON-NLS-1$
                                                 processMonitorThread = new Thread(psMonitor);
                                                 processMonitorThread.start();
+                                                // Local link for testing ESB related jobs
+                                                addEndpointURL();
```
sbouchet: see at the end of the class about this new method. ( adding a message only for esb )

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/cmd/ChangeValuesFromRepository.java
nrousseau: Seems a little strange... but indeed, pass the node to the drag&drop handler would need many refactor.
I don't think there is more simple for now...

main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/RunProcessContext.java
nrousseau: Note that if you check in the emf model, it will only cover the current job, and not the components included in joblets.
Is it expected or missing? as i think ESB components could also be in any joblet as well.

What about the case also of tRunJob? Only the mainjob will show this? no need for children jobs?

sbouchet: Good point. this first implementation is only focusing on current job/route. i need feedback from PM to know if we handle also tRunJob and joblets.

PR #3025, opened 2018-11-29
fix(TDI-41404): Avoid 64kb method issue when too many context paramet…

dmytro-sylaiev:
…er exists

64kB issue exist when a lot of context variables declared
**What is the new behavior?**
Inner class with split-methods used

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:


dmytro-sylaiev:
@wwang-talend 
It was the first idea to handle this issue. 
It is ok for tRunJob code, but, for `tRunJob_main.javajet` we still need to declare local class since generated code is inside cidProcess() method.  But it'd still be OK.
Another pb is for footer (part of handling current jobs' context variables in runJobInTos).
There we generate next code to set context values `context.<%=ctxParam.getName()%> = someValue` therefore we can't just extract it to some external method.


So I thought that it's great opportunity to support jobs with any amount of context variables and fix customers pb simultaneously.

PR #3215, opened 2019-01-31
fix(TDI-37349): Join thread in the finally block

dmytroshekera:
* Handle error and quick exit with catch-finally block
  but child thread proceed working so wait for it end

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/tWriteXMLFieldOut/tWriteXMLFieldOut_begin.javajet
+++ b/main/plugins/org.talend.designer.components.localprovider/components/tWriteXMLFieldOut/tWriteXMLFieldOut_begin.javajet
@@ -157,6 +157,7 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
 					<%} else {%>
 						ThreadXMLField_<%=cid%> txf_<%=cid%> = new ThreadXMLField_<%=cid%>(listGroupby_<%=cid%>, flows_<%=cid%>);
 					<%}%>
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.components.localprovider/components/tWriteXMLFieldOut/tWriteXMLFieldOut_begin.javajet
```diff
@@ -157,6 +157,7 @@ if ((metadatas!=null)&&(metadatas.size()>0)) {
 					<%} else {%>
 						ThreadXMLField_<%=cid%> txf_<%=cid%> = new ThreadXMLField_<%=cid%>(listGroupby_<%=cid%>, flows_<%=cid%>);
 					<%}%>
+					globalMap.put("wrtXMLFieldIn_<%=cid%>", txf_<%=cid%>);
```
gonchar-ivan: By the way I found that for other resource (e.g. queue_<%=virtualTargetCid%>) virtualTargetCid is used.
As I understand the difference is following:
<%Cid%> = tWriteXMLField_1_Out
<%=virtualTargetCid%> = tWriteXMLField_1_In
The first variable holds a name of current component. The second variable holds the name of next component.
(I just had a question why virtualTargetCid is also used. However, I think that both variables are ok)

gonchar-ivan: The only think I suggest to rename key from `wrtXMLFieldIn_<%=cid%>` to `thread_<%=cid%>` emphasize that a thread is stored under this key.
But it's also ok for me to leave without changes

PR #3344, opened 2019-03-25
TDQ-16354 TOP i18n preference page

xqliu-talend:


xqliu-talend:
> at least broken on
> 
> [tdi-studio-se/main/plugins/org.talend.designer.core/plugin.xml](https://github.com/Talend/tdi-studio-se/blob/0f7a91438599d56448d7bb0be2601eff4a910caf/main/plugins/org.talend.designer.core/plugin.xml#L90)
> 
> Line 90 in [0f7a914](/Talend/tdi-studio-se/commit/0f7a91438599d56448d7bb0be2601eff4a910caf)
> 
>  class="org.talend.designer.core.ui.preferences.I18nPreferencePage" 
> as it won't find the class

actually, the class "org.talend.designer.core.ui.preferences.I18nPreferencePage" still exist, I create a new abstract class "org.talend.core.ui.preference.I18nPreferencePage" in org.talend.core.ui, both "org.talend.designer.core.ui.preferences.I18nPreferencePage" and "org.talend.dataprofiler.rcp.i18n.I18nPreferencePage" extend from it, so the TDQEE and TOP have the same internationalization preference page

PR #3410, opened 2019-04-16
TESB-25571 activemq-client version is upgraded and excluded from generated OSGi bundle

apoltavtsev:


wwang-talend:
should not put to patch/7.1.1 as it works only for TUP patch, not for the component plain patch. And if want to put patch/7.1.1, need TUP team agree.

wwang-talend:
have create another pull request and merged to master branch, so can close

PR #3512, opened 2019-05-24
Hliashchuk/fix/tbd 8649

hliashchuk:

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:


main/plugins/org.talend.designer.components.localprovider/components/tBigQueryInput/tBigQueryInput_begin.javajet
```diff
@@ -230,15 +231,32 @@
...
 				.getService();
-				
+
+		<%
+		if (resultSizeType.equals("LARGE") || resultSizeType.equals("AUTO")) {
+		%>
 		com.google.cloud.bigquery.QueryJobConfiguration queryConfiguration_<%=cid%> = 
 																com.google.cloud.bigquery.QueryJobConfiguration.newBuilder(<%=query%>)
 																.setUseLegacySql(<%=useLegacySql%>)
 																.setDestinationTable(com.google.cloud.bigquery.TableId.of(dataset, temp_table))
```
lbourgeois: Seems there is code duplication here, I think only lines 241-243 should be put in a if block 

Code formatting issue

main/plugins/org.talend.designer.components.localprovider/components/tBigQueryInput/tBigQueryInput_begin.javajet
```diff
@@ -249,18 +267,6 @@
 		%>
 		job_<%=cid%> = job_<%=cid%>.waitFor();
 		
-		if (job_<%=cid%>.getStatus().getError() != null && "responseTooLarge".equals(job_<%=cid%>.getStatus().getError().getReason()) && "AUTO".equals("<%=resultSizeType%>")) {
```
lbourgeois: Why having removed this part ? Looks the correct way to implement AUTO mechanism (first try without allow large result flag and retry with iin case of error)

hliashchuk: From what I've experienced, you'll get the "response too large" error after processing near 15 GB of data. And it took me around 8 hours, so I thought it would be better to process it on a first try. In case if the actual dataset can be treated as "SMALL" and a customer will pick "AUTO", there is no such huge difference in the processing time, data is temporarily stored in a table which is deleted afterwards.

main/plugins/org.talend.designer.components.localprovider/components/tBigQueryInput/tBigQueryInput_begin.javajet
hliashchuk: Thanks, will correct according to your note

PR #3521, opened 2019-05-29
fix(TUP-21073): [GuessSchema] Guess Schema process fails in case of messages in error stream

cmeng-talend:

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:
https://jira.talendforge.org/browse/TUP-21073


main/plugins/org.talend.sdk.component.studio-integration/messages.properties
```diff
@@ -24,6 +24,7 @@ TaCoKitConfiguration.wizard.exception.message.default=Please see details below.
 TaCoKitConfiguration.composite.exception.message.default=An error happened during the operation
 guessSchema.button=Guess {0} Schema
 guessSchema.error=Error ocurred during guessing schema.
+guessSchema.error.empty=Guess schema failed, no error message returned from the guess schema execution.
```
akhabali: This error message is useless

cmeng-talend:
I think we can't judge guess schema succeed or not though empty status of error stream, cause we can't make sure users' component won't print anything into error stream during their component execution, we should judge guess schema succeed or not based on the execution result instead of the error stream:
If returnCode == 0, means guess schema succeed, we can try to get result from out stream;
If returnCode != 0, means guess schema failed, we can read the error stream and show them to users.

cmeng-talend:
@akhabali how do you think?

rmannibucau:
Guess the opposite works better: if you got expected json line - we can ignore others - from stdout then all good, else read stderr and show it all. Exit code = 0 does not help as you explained.

cmeng-talend:
@rmannibucau updated

cmeng-talend:
@rmannibucau fix updated, since we need to check all lines starting with '[' and ending with ']', and there may be risk those lines are logs printed by user's component, so we have to get those lines and error stream at the same time when resolving metadata from those lines, if no valid metadata resolved, then means maybe guess schema failed, will print all messages from error stream.

main/plugins/org.talend.sdk.component.studio-integration/messages.properties
cmeng-talend: will still print it if no any error returned from guess schema job, it's more friendly

main/plugins/org.talend.sdk.component.studio-integration/src/main/java/org/talend/sdk/component/studio/ui/guessschema/GuessSchemaSelectionAdapter.java
```diff
@@ -179,31 +197,42 @@ public void widgetSelected(final SelectionEvent event) {
     }
 
     private IMetadataTable buildMetadata(final String schema) {
-        if (schema.trim().isEmpty()) {
+        if (schema == null || schema.trim().isEmpty()) {
             return null;
         }
+        Collection<MetadataColumn> jsonColumns = new HashSet<>();
+        final Jsonb jsonb = JsonbBuilder.create();
```
rmannibucau: should be a try with close resource as it was before

cmeng-talend: @rmannibucau ok, updated fix, thanks

cmeng-talend:
Hello @rmannibucau  @akhabali , does this PR have other issue? if there is no issue, could we merge it now? we need to merge this fix and test it as soon as possible for the release.

PR #3654, opened 2019-06-24
TESB-26300: disable message dialog when checking build errors

sbouchet:
When route is built, it shows twice an error dialog.

**What is the new behavior?**
Do not display dialog on error check

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/JobErrorsChecker.java
+++ b/main/plugins/org.talend.designer.runprocess/src/main/java/org/talend/designer/runprocess/JobErrorsChecker.java
@@ -31,7 +31,6 @@ import org.talend.commons.exception.CommonExceptionHandler;
 import org.talend.commons.exception.ExceptionHandler;
 import org.talend.commons.exception.PersistenceException;
 import org.talend.commons.exception.SystemException;
-import org.talend.commons.ui.runtime.exception.MessageBoxExceptionHandler;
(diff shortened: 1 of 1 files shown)


nrousseau:
strange to be linked to task TESB-23600

For me the most important is to:
- have a way to see there is a compilation error when execute a job (DI/BD)/route from studio (not sure it's the case now?)
- have a popup of error still when build a job.

Agreed that the MessageBoxExceptionHandler in the middle of the build was a bit too much... as long as exception is catched / message displayed in another place.

Just a note.. MessageBoxExceptionHandler already had a case to avoid the popup of dialog

sbouchet:
@nrousseau 
Yes, i'm also a bit confused about the dialog in the middle of the process of the build. 
tried to reproduce on the custom build but the master branch does not seems to work fine, a lot of error due to some dependencies still on 7.2.1 and so. 
I agree to check on the DI job side that this is not causing any regression, as far as i tested, it seems not. since the custom build does not trigger anymore the junit, i cannot tell. 
Also, the strange behavior is that this error check is performed TWICE when building the route, the stack shows that the export of the route is made also TWICE ... the complexity of the route building is something to look at eventually (cc @dsergent )

dsergent:
TESB-23600 is a mistake. I updated the PR title to reflect the appropriate jira: TESB-26300.

PR #3697, opened 2019-07-10
fix(TUP-23729):Memory issue found when run TUJ (#3689)

zwxue:


zwxue:
won't close, may need later

PR #4202, opened 2019-12-05
TUP-25343 [BUG] database can't dataview after set routine.encryption.key.v2

kjwang-talend:
TUP-25343 [BUG] database can't dataview after set routine.encryption.key.v2
https://jira.talendforge.org/browse/TUP-25343

**What is the new behavior?**

- [X] Bugfix
- [ ] Feature

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/runprocess/Processor.java
+++ b/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/runprocess/Processor.java
@@ -389,16 +389,26 @@ public abstract class Processor implements IProcessor, IEclipseProcessor, Talend
     protected String[] addEncryptionFilePathParameter(String[] cmds) {
         String encryptionFilePath = System.getProperty(StudioKeysFileCheck.ENCRYPTION_KEY_FILE_SYS_PROP);
         File encryptionFile = new File(encryptionFilePath);
-        List<String> cmdList = new ArrayList<String>();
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/runprocess/Processor.java
```diff
@@ -381,6 +386,21 @@ protected Process exec(String[] cmd, String path) throws ProcessorException {
...
+        for (int i = 0; i < cpIndex; i++) {
+            cmdList.add(cmds[i]);
+        }
+        cmdList.add(StudioKeysFileCheck.ENCRYPTION_KEY_FILE_JVM_PARAM + "=" + encryptionFile.toURI().getPath());
+        for (int i = cpIndex; i < cmds.length; i++) {
+            cmdList.add(cmds[i]);
+        }
+        return cmdList.toArray(new String[0]);
+    }
+
```
wchen-talend: This function seems will always add the ENCRYPTION_KEY_FILE_JVM_PARAM  to the cmdList ?
Can we add it only if this argument is not set from job ?
I'm afraid it will overwrite the setting in job .

wchen-talend: Ignore my last comment if the function is only used in local run .
It should be no problem job always use the same key file as studio to run .

PR #4491, opened 2020-02-27
TDI-43764 : MsSQL NVARCHAR

rlecomte-talend:
Use setNString method for NVARCHAR and NCHAR columns

Method setString il always used to set the value of a string column

**What is the new behavior?**
Code now checks if the column is a NVARCHAR or NCHAR type and then use setString or setNString accordingly

- [X ] Bugfix
- [ ] Feature

- [ ] Yes
- [ X] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
+++ b/main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
@@ -2062,7 +2062,13 @@ public class CLASS {
                     setStmt.append(prefix + cid + ".setNull(" + index + dynamic + ", java.sql.Types.DATE);\r\n");
                 }
                 setStmt.append("}\r\n");
-            } else {
+            } else if(typeToGenerate.equals("String")) {
+                   if (column.dataType.equalsIgnoreCase("NVARCHAR") || column.dataType.equalsIgnoreCase("NCHAR")) {
+                        setStmt.append(prefix + cid + ".setNString(" + index + dynamic + ", " + incomingConnName + "." + column.getName() + ");\r\n");
+                   } else {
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
```diff
@@ -2062,7 +2062,13 @@ public class CLASS {
                     setStmt.append(prefix + cid + ".setNull(" + index + dynamic + ", java.sql.Types.DATE);\r\n");
                 }
                 setStmt.append("}\r\n");
-            } else {
+            } else if(typeToGenerate.equals("String")) {
```
clesaec: Like in the code upper, use the style "constant.equals(var)"
```
if("String".equals(typeToGenerate) {
if ("NVARCHAR".equalsIgnoreCase(column.dataType) || 
```
It won't fail if a value is null (and it's hard to see if it can be or not with the function size).

rlecomte-talend: I tested "String " type the same way as Date and Char/Character types are tested a few lines above. (cf lines 2007 and 2019).

clesaec: What i read on line 2059 is `if ("TIME".equals(dbType)) {`
(is it not better to correct line 2007 & 2019 ?) typeToGenerate var is input in function call, nothing tell that it can't be null.

rlecomte-talend: It is the not the same variable that is checked (dbType vs typeToGenerate). "typeToGenerate" is always tested as I did it (and that's why I did it this way). And I won't "correct" something that isn't the scope of the issue.

PR #4556, opened 2020-03-25
TUP-26609 - Harden deserialization for codegen

coheigea:
This task is to harden deserialization in tdi-studio-se/main/plugins/org.talend.designer.codegen, to make sure that we are only deserializing what we are expecting.

Note: I can't build the code, so it may have compilation errors - it needs to be tested also.

### DIFF
--- a/main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/model/JetSkeletonManager.java
+++ b/main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/model/JetSkeletonManager.java
@@ -20,8 +20,10 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InvalidClassException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/model/JetSkeletonManager.java
```diff
@@ -170,7 +172,18 @@ private void deserializeAlreadyChecked() throws Exception {
         BufferedInputStream bufferedInputStream = null;
         try {
             bufferedInputStream = new BufferedInputStream(new FileInputStream(file));
-            ObjectInputStream objectIn = new ObjectInputStream(bufferedInputStream);
+            ObjectInputStream objectIn = new ObjectInputStream(bufferedInputStream) {
+                @Override
+                protected Class<?> resolveClass(ObjectStreamClass desc) throws IOException, ClassNotFoundException {
+                    // Enforce the types that we are about to deserialize
+                    if (!(desc.getName().startsWith("java.util")
```
wwang-talend, line 179: the package "java.util.iambadguy" is possible?

wwang-talend: then can skip the security check

wwang-talend (approved):

main/plugins/org.talend.designer.codegen/src/main/java/org/talend/designer/codegen/model/JetSkeletonManager.java
coheigea: In theory yes, but remember the package has to be on the classpath of the application, so it's unlikely that the application will be allowing this package ;-)

wwang-talend (approved):

nrousseau:
seems code was re-written... not sure why... closing the PR since it's done

PR #4643, opened 2020-04-27
fix(TDI-44129):  Spatial / PostGIS / Error on GeomFromText

fxprunayre:
Fix for:
* https://github.com/talend-spatial/talend-spatial/issues/82
* https://jira.talendforge.org/browse/TDI-44129

See https://postgis.net/docs/reference.html

```
	
PostGIS has begun a transition from the existing naming convention to an SQL-MM-centric convention. As a result, most of the functions that you know and love have been renamed using the standard spatial type (ST) prefix. Previous functions are still available, though are not listed in this document where updated functions are equivalent. The non ST_ functions not listed in this documentation are deprecated and will be removed in a future release so STOP USING THEM.
```

**What is the current behavior?** 

DB exception

**What is the new behavior?**

No exception

- [x] Bugfix

- [x] No


fxprunayre:
Closing as duplicates of https://github.com/Talend/tdi-studio-se/pull/1320

PR #4651, opened 2020-04-29
TDI-43935 : MsSQLRow NString type for prepared statement

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-43935

MsSQLDbRow does not support NString as prepared statement parameter type

**What is the new behavior?**
NString is added in supported types list


main/plugins/org.talend.designer.components.localprovider/components/tMSSqlRow/tMSSqlRow_java.xml
```diff
@@ -298,6 +298,7 @@
         		<ITEM NAME="Object" 		VALUE="Object"/>
         		<ITEM NAME="Short" 			VALUE="Short"/>
         		<ITEM NAME="String" 		VALUE="String"/>
+                <ITEM NAME="NString"        VALUE="NString"/>
```
ypiel-talend, line 301: Strange, this lis is a java type list. NString is a SQL type.

ypiel-talend: Ok it is the end of the jdbc method getXxxx : https://github.com/Talend/tdi-studio-se/blob/cb8285f0c69085cf038e408b0514bda39fcd8f12/main/plugins/org.talend.designer.components.localprovider/components/templates/DB/Row/AbstractDBRowMain.javajet#L107

rlecomte-talend: Yes, not the cleanest way to do it, but it works.

PR #4804, opened 2020-06-12
TDI-44374 - Update BeanUtils to 1.9.4 in talend-ws

coheigea:
Veracode is reporting a high level CVE against BeanShell 1.8.3 in:

tdi-studio-se/main/plugins/org.talend.designer.components.libs/libs_src/talend-ws
This task is to update it to 1.9.4.


coheigea:
Fixed by a different PR.

PR #4910, opened 2020-07-09
TESB-29188 A Dataservice job is not working when deployed to a server

msheverda-talend:
The "\\" character should be escapted in feature.xml file when publish the Services to TMC.
More details in last comment here https://jira.talendforge.org/browse/TESB-29188


andreasmattes:
The present issue will be handled globally in the scope of https://jira.talendforge.org/browse/TPSVC-14902 - therefore we close the present pull request in favour of their solution.

PR #4929, opened 2020-07-14
TESB-29647 Bundle version does not use 'Custom Version' for jobs

msheverda-talend:
If the job's deployment "Custom Version" is different than that of the job's item version, then the job's bundle uses the job's item version instead of "Custom Version". This PR is to make bundle use "Custom Version"


dsergent:
No PR needed for this branch.

PR #5012, opened 2020-07-30
tS3Connection setPathStyleAccess for better MinIO support

bgunics-talend:
Added a checkbox under Advanced settings of tS3Connection when a custom endpoint is used to enable pathstyle access.

When MinIO is used via DNS/hostname this must be checked otherwise the operations fail.

https://jira.talendforge.org/browse/TDI-44045

**What is the new behavior?**
https://jira.talendforge.org/browse/TDI-44045

- [ ] Bugfix
- [X] Feature

- [ ] Yes
- [X] No

**Other information**:


dmytro-sylaiev:
Hello @bgunics-talend 
You have added checkbox only for tS3Connection, but I think also need to add it for every tS3XXX component when not use the existing connection
I'll do that in separate PR

bgunics-talend:
Hello @dmytro-sylaiev  unfortunately that would require a lot of change in all the other components. The custom region endpoint seems to be only available in the tS3Connection component. 
For example compare the tS3Get and tS3Connection advanced setting.

main/plugins/org.talend.designer.components.localprovider/components/tS3Connection/S3Client.javajet
```diff
@@ -408,6 +412,9 @@
                     <%}%>
                     .withCredentials(credentialsProvider_<%=cid%>)
                     .withClientConfiguration(cc_<%=cid%>)
+					<%if(enablePathStyleAccess) {%>
+					.enablePathStyleAccess()
```
jzhao-talend, line 416: When **USE_REGION_ENDPOINT** is **unchecked** and **PATHSTYLEACCESS** is **checked**.
This should not be generated. but currently I think the **.enablePathStyleAccess()** would be generated.
Please check this.

bgunics-talend: I've added the useRegionEndpoint this way PathStyle is ignored if the custom region is not set.

PR #5062, opened 2020-08-11
TDI-44682 - Fix some locale issues

coheigea:
https://jira.talendforge.org/browse/TDI-44682


main/plugins/org.talend.designer.components.libs/libs_src/talend-mscrm/src/main/java/org/talend/ms/crm/sdk/DeviceIdManager.java
```diff
@@ -264,12 +264,12 @@ private static String parseRegistrationResponse(HttpResponse response, Boolean i
                 }
 
                 // Get the puid and return it.
-                if (node.getNodeName().toUpperCase().equals("PUID")) {
+                if (node.getNodeName().equalsIgnoreCase("PUID")) {
```
clesaec, line 267: Even better, when there's a constant, is to write constant.equalsIgnoreCase(variable)
Here
`if ("PUID".equalsIgnoreCase(node.getNodeName())) {`
Won't fail if getNodeName() return null.

coheigea: @clesaec Done, thanks.

PR #5065, opened 2020-08-11
TDI-44685 - Close streams after calling getResourceAsStream in genera…

coheigea:
https://jira.talendforge.org/browse/TDI-44685


wwang-talend (approved):
LGTM

PR #5042, opened 2020-08-06
fix(TDI-44568): Update org.talend.libraries.excel plugin

dmytrogrygorenko:
**What is the current behavior?** (You can also link to an open issue here)

**What is the new behavior?**

**Please check if the PR fulfills these requirements**

- [X] The commit message follows Talend standard
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features) ?
- [ ] The code coverage on new code >75%
- [X] The new code does not introduce new technical issues (sonar / eslint)

**What kind of change does this PR introduce?**

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build / CI related changes
- [ ] Other... Please describe:

**Does this PR introduce a breaking change?**

- [ ] Yes
- [X] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

PR #5251, opened 2020-09-22
Revert "fix(TDI-44568): Update org.talend.libraries.excel plugin"

dmytrogrygorenko:
Reverts Talend/tdi-studio-se#5042


undx (approved):
LGTM

PR #5289, opened 2020-09-29
fix(TDI-44935): reuse custom query expression

ozhelezniak-talend:
* make query expression idempotent in the generated code
* use instead of duplication of custom query expression a variable
* store the expression in the variable and then read it again
* make obsolete the 'query' method with one argument

**What is the current behavior?** (You can also link to an open issue here)
https://jira.talendforge.org/browse/TDI-44935

**What is the new behavior?**
invoke query expression only once that prevents the change of logic for non-idempotent queries.

**Please check if the PR fulfills these requirements**

- [x] The commit message follows Talend standard

**What kind of change does this PR introduce?**

- [x] Bugfix

**Does this PR introduce a breaking change?**

- [x] May. It depends on the customer's usage of the logger.

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

PR #5349, opened 2020-10-13
fix(TDI-44935): reuse custom query expression

ozhelezniak-talend:
* make query expression idempotent in the generated code
* use instead of duplication of custom query expression a variable
* store the expression in the variable and then read it again
* make obsolete the 'query' method with one argument

**What is the current behavior? (You can also link to an open issue here)**
https://jira.talendforge.org/browse/TDI-44935

**What is the new behavior?**
invoke query expression only once that prevents the change of logic for non-idempotent queries.

- [x] Bugfix

- [x] May. It depends on the customer's usage of the logger.

**Other information**:


ozhelezniak-talend:
It's a twin brother of closed https://github.com/Talend/tdi-studio-se/pull/5289 (after rebasing it's become a mess and it was hard to tell what was changed).

Voievidko (approved):
LGTM

Voievidko (approved):
LGTM

PR #5396, opened 2020-10-20
fix(TDI-44935): reuse custom query expression

ozhelezniak-talend:
* make query expression idempotent in the generated code
* store the expression in the variable and then read it again

https://jira.talendforge.org/browse/TDI-44935

**What is the new behavior?**
the query is idempotent for logger

- [x] Bugfix

- [x] No

**Other information**:
It's a backport from maintenance/7.3


Voievidko (approved):
LGTM

PR #5502, opened 2020-11-12
Rlecomte/tdi 45147 t gs copy target issue

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-45147

When a blog is copied from one bucket to another, its orginal path is used as the new path  aven totally ; ingoring new path when service account is used

**What is the new behavior?**
The blog is now copied where the user wants it


dmytro-sylaiev (approved):
LGTM

PR #5516, opened 2020-11-16
fix(TESB-29553) Publishing a route with cTalendJob from Studio and commandline gives different results

apoltavtsev:


dsergent (approved):
LGTM

PR #5521, opened 2020-11-17
TDI-45007 / TDI-45144 / TDI-31602 GPLoad BugFix / enhancements

bgunics-talend:
TDI-45007 / TDI-45144 / TDI-31602 GPLoad BugFix / enhancements

**What is the new behavior?**

- [X] The new code does not introduce new technical issues (sonar / eslint)

- [X] Bugfix
- [X] Feature
- [X] Code style update (formatting, local variables)

- [ ] Yes
- [X] No

**Other information**:

## tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet / tGreenplumGPLoad/tGgreenplumGPLoad_java.xml

Added guessColumnList so the yaml will contain the columns of the talend schema, this is extremely useful if the target table has extra columns (for example to load only the primary keys to a staging table) This will generate the "  COLUMNS:" section of the YAML file.

Added parameters to the advanced settings, more information for these can be found at: https://gpdb.docs.pivotal.io/6-12/utility_guide/ref/gpload.html#topic1__section7
Parameters include:
  LOG_ERRORS:
  MAX_LINE_LENGTH:
  EXTERNAL:
    -SCHEMA:
  PRELOAD:
    -TRUNCATE:
    -REUSE_TABLES:
    -STAGING_TABLE:
    -FAST_MATCH:
  SQL:
    -BEFORE:
    -AFTER:

Changed the way command is built, so the Log4j now displays a command that can be copy-pasted to the console for testing.
The yaml file was missing from it.
The PythonPath was also not printed to the log4j logs.

Sometimes a database connection was opened and closed without executing any actions.

Password was never passed to gpload utility so it had to be managed separately, it is now passed via context variable, which means when executed by talend it won't prompt for a password and when executing from console it will prompt for a password.

GPLoad will terminate if there was an error (for example target table does not exists)

Extra after variables, for the gpload details:
		NB_LINE_INSERTED	Integer
		NB_LINE_UPDATED		Integer
		NB_DATA_ERRORS		Integer
		STATUS				String
		RUNTIME				String

Option to GZip compress the datafile, this will increase CPU usage, but decreases DISK usage. (50-90% smaller files)
Remove data file after successful execution. If there were *any* errors the data file is not removed. 

## tGreenplumGPLoad/tGreenplumGPLoad_main.javajet
Added byte[] type support. Byte[] is converted to escaped OCTET characters, which is compatible with all PostgreSQL versions from 8 to 13.

Fixed when newline / escape characters appeared in the data.

Postgresql uses null terminated bytes so those have to be removed from the data to avoid errors.

## tGreenplumGPLoad/tGreenplumGPLoad_end.javajet
BugFix don't overwrite the actual inserted record number with 0


main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
```diff
@@ -97,9 +99,19 @@ skeleton="../templates/db_output_bulk.skeleton"
...
 	boolean isLog4jEnabled = ("true").equals(ElementParameterParser.getValue(node.getProcess(), "__LOG4J_ACTIVATE__"));
 	
+%>
+	String dataFile_<%=cid%> = <%=dataFile%>;
+<%
+	if(useGzip) {
+		//make sure it ends with .gz
+%>
+	if(!dataFile_<%=cid%>.endsWith(".gz")) {
+			dataFile_<%=cid%> += ".gz";
```
jzhao-talend, line 109: I don't think we need to change the file name during runtime.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
```diff
@@ -468,6 +482,45 @@ skeleton="../templates/db_output_bulk.skeleton"
...
+				  IMetadataTable metadataTable = metadataList.get(0);
+				  if (metadataTable != null) {
+				  	java.util.HashMap <String,String> columnTypeHashMap = new java.util.HashMap<String,String>();
+				  	columnTypeHashMap.put("id_String","text");
+				  	columnTypeHashMap.put("id_byte[]","bytea");
+				  	columnTypeHashMap.put("id_Date","timestamp");
+				  	columnTypeHashMap.put("id_BigDecimal","numeric (500,100)");
+				  	columnTypeHashMap.put("id_Double","numeric (500,100)");
+				  	columnTypeHashMap.put("id_Integer","integer");
+				  	for (IMetadataColumn metadataColumn : metadataTable.getListColumns()) {
```
jzhao-talend, line 517: this only use the fixed type mapping. i think we should can use the db type in the schema

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
```diff
@@ -482,6 +535,23 @@ skeleton="../templates/db_output_bulk.skeleton"
 			gpctrl_<%=cid%>.println("    - HEADER: true");
 			<% } %>
 			gpctrl_<%=cid%>.println("    - ENCODING: " + <%=encoding%>);
+
+			<%
+			for (Map<String, String> option : options) {
```
jzhao-talend, line 540: this part of code should be improve, as it only for "EXTERNAL_SCHEMA"

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
```diff
@@ -517,13 +587,66 @@ skeleton="../templates/db_output_bulk.skeleton"
 					<%
 				}
 			}
+
+			List<String> preload = new ArrayList<String>();
+			for (Map<String, String> option : options) {
```
jzhao-talend, line 592: this part too.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
```diff
@@ -569,34 +692,63 @@ skeleton="../templates/db_output_bulk.skeleton"
...
 				java.io.BufferedReader reader = new java.io.BufferedReader(new java.io.InputStreamReader(ps.getInputStream()));
 				java.io.BufferedReader errorReader = new java.io.BufferedReader(new java.io.InputStreamReader(ps.getErrorStream()));
 				String line = "";
 				while ((line = reader.readLine()) != null) {
-					gploadOutput_<%=cid%>.append(line);
-					System.out.println(line);
-					if (line.contains("rows Inserted")) {
-						nbLine_<%=cid %> = getNBLine(line);
-					}
+					gploadOutput_<%=cid%>.append(line+"\n");
```
jzhao-talend, line 738: this would change the globalMap var GPLOAD_OUTPUT value, not sure be accept.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_messages.properties
```diff
@@ -82,3 +91,14 @@ SPECIFY_GPLOADPY_PATH.NAME=Specify gpload.py path
...
 DIE_ON_ERROR.NAME=Die on error
+
+
+NB_LINE_INSERTED.NAME=NB_LINE_INSERTED
+NB_LINE_UPDATED.NAME=NB_LINE_UPDATED
+NB_DATA_ERRORS.NAME=NB_DATA_ERRORS
+STATUS.NAME=GPLOAD_STATUS
+RUNTIME.NAME=GPLOAD_RUNTIME
+ERRORMESSAGE.NAME=GPLOAD_ERRORMESSAGE
+GUESS_COLUMN_LIST.NAME=Populate Column List Based on schema
```
jzhao-talend, line 102: need follow our name rule of parameter

jzhao-talend:
Thanks @bgunics-talend for your contribution.
I would cherry-pick your commits, and do some update based on. then create a new PR follow our workflow.
would keep you as a contributor.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
bgunics-talend: If it doesn't ends with GZ and we place compressed data it will error out. As gpfdist only looks for the .gz extension.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
bgunics-talend: It should be yes, but my problem was that if I try to mix this with some other parts it might results in a non well formatted yaml file. So this section has to be on its own, else having the advanced parameters ordered in a wrong way could cause yaml errors.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
bgunics-talend: Does it look slightly better this way?

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
jzhao-talend: this seems only work for tGreenplumnGPLoad with input.
when it's a  standalone component the usezip would not work with it.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
jzhao-talend: @bgunics-talend do you have the document about this kind of maping ? 
like **id_BigDecima : numeric (500,100)**

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
jzhao-talend: i means at least should not cotninue loop when get **EXTERNAL_SCHEMA**

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
```diff
@@ -569,34 +683,72 @@ skeleton="../templates/db_output_bulk.skeleton"
...
+						} else if(line.contains("|WARN|")){
+							log.warn("<%=cid%> - gpload says: "+ line);
+						} else if(line.contains("|INFO|")){
+							log.info("<%=cid%> - gpload says: "+ line);
+						} else {
+							log.debug("<%=cid%> - gpload says: "+ line);
+						}
+					<%}
+						if(!isLog4jEnabled){%>
+						System.out.println(line);
```
jzhao-talend, line 742: we should not use **System.out**, when customer want to see this kind of issue we should suggest them active log4j

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_main.javajet
```diff
@@ -158,7 +189,9 @@ skeleton="../templates/db_output_bulk.skeleton"
 					temp_<%=cid%> = String.valueOf(<%=inputRowName%>.<%=column.getLabel()%>);
 					temp_<%=cid%> = temp_<%=cid%>.replace(<%=escapeChar%>,escapeCharRepl_<%=cid%>);
 					temp_<%=cid%> = temp_<%=cid%>.replace(<%=quoteChar%>,quoteCharRepl_<%=cid%>);
-					temp_<%=cid%> = temp_<%=cid%>.replace("\n",rowSepRepl_<%=cid%>);
+					temp_<%=cid%> = temp_<%=cid%>.replaceAll("\\\n", "\\\\012").replaceAll("\\\r","\\\\015");
```
jzhao-talend, line 192: after this change it would be change the colum value which insert into table:
image

jzhao-talend: what's the case about csv mode ? i see your tuj can work without this part of fix.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
bgunics-talend: This is to keep the old behavior where everything was sent to system out. :/ 
Then I added log4j, but it was by default debug, which filtered out too much information, so I was thinking about maybe adding a checkbox (like in TDI-31602 to force display of logs to console, but the gpload -> log4j rewrite with keeping the levels seems to do the trick, so scrubbed that idea)

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_main.javajet
bgunics-talend: In linux we only have \n in windows we have a combination of \n and \r.  If you check the byte[] part, there we also used this octal escaped string. This has 2 results:

-  you no longer have \\\n but have \012 so the file changes 👎 
- however now each line represents a record, so when someone does a wc -l then the correct number of lines will be reported. (Extremely helpful when you have 100+ GB of data to load.) 👍 

It's for ease of use.

jzhao-talend: the target table content is different, this would break regression.
so we can't accept this part of fix.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
bgunics-talend: Unfortunately I don't, the reason for using such values were mostly based on trial & error at the time. (When we don't have the right length / precision filled in the component this way it doesn't error out.) Yes we could use the values from the schema itself, however based on my experimentation this is only used for the external table definition.

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_main.javajet
bgunics-talend: Removed this part of the code

main/plugins/org.talend.designer.components.localprovider/components/tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet
bgunics-talend: Added an extra check for hasInputRow.

bgunics-talend:
Since this pull request has been cherry picked and implemented by @jzhao-talend  I'll close this Pull Request

PR #5627, opened 2020-12-14
fix(TUP-29693): Fix migrating jarname to maven uri if jar name contai…

bhe-talendbj:
…ns context

https://jira.talendforge.org/browse/TUP-29693

**What is the new behavior?**
1 - If jarname contains context, then migrate the value as it is, otherwise replace value with maven uri.
2 - If wrong maven uri was set because of R2020-11 patch, then reset the maven uri back to context.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/utils/UpdateModuleListInComponentsMigrationTask.java
+++ b/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/utils/UpdateModuleListInComponentsMigrationTask.java
@@ -21,10 +21,13 @@ import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.emf.common.util.EList;
 import org.talend.commons.exception.ExceptionHandler;
+import org.talend.commons.exception.PersistenceException;
 import org.talend.commons.runtime.model.emf.EmfHelper;
 import org.talend.core.CorePlugin;
+import org.talend.core.GlobalServiceRegister;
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/utils/UpdateModuleListInComponentsMigrationTask.java
```diff
@@ -69,6 +80,73 @@
...
+            for (IRepositoryViewObject object : list) {
+                ExecutionResult execute = null;
+                // in case the resource has been modified (see MergeTosMetadataMigrationTask for example)
+                if ((object.getProperty().eResource() == null || object.getProperty().getItem().eResource() == null)
+                        && (object instanceof RepositoryObject)) {
+                    Property updatedProperty = factory.reload(object.getProperty());
+                    ((RepositoryObject) object).setProperty(updatedProperty);
+                }
+
+                execute = this.migrateProjectSettings(project);
```
wchen-talend, line 112: why need execute migrateProjectSettings in the loop ? 
It will try to check the project migration many times . It should be a one time check migration.

wchen-talend: I think UpdateModuleListInComponentsMigrationTask.java  should keep the main logic the same as before . but only add the check for context mode in module list .

dlin-talend (approved):

PR #5709, opened 2021-01-06
feat(TDI-45007 & TDI-31602): Enhance component tGreenplumGPLoad 

jzhao-talend:

for jira:
https://jira.talendforge.org/browse/TDI-31602 
https://jira.talendforge.org/browse/TDI-45007

**What is the new behavior?**

## tGreenplumGPLoad/tGreenplumGPLoad_begin.javajet /
tGreenplumGPLoad/tGgreenplumGPLoad_java.xml

Added guessColumnList so the yaml will contain the columns of the talend
schema, this is extremely useful if the target table has extra columns
(for example to load only the primary keys to a staging table) This will
generate the "  COLUMNS:" section of the YAML file.

Added parameters to the advanced settings, more information for these
can be found at:
https://gpdb.docs.pivotal.io/6-12/utility_guide/ref/gpload.html#topic1__section7
Parameters include:
  LOG_ERRORS:
  MAX_LINE_LENGTH:
  EXTERNAL:
    -SCHEMA:
  PRELOAD:
    -TRUNCATE:
    -REUSE_TABLES:
    -STAGING_TABLE:
    -FAST_MATCH:
  SQL:
    -BEFORE:
    -AFTER:

Changed the way command is built, so the Log4j now displays a command
that can be copy-pasted to the console for testing.
The yaml file was missing from it.
The PythonPath was also not printed to the log4j logs.

Sometimes a database connection was opened and closed without executing
any actions.

Password was never passed to gpload utility so it had to be managed
separately, it is now passed via context variable, which means when
executed by talend it won't prompt for a password and when executing
from console it will prompt for a password.

GPLoad will terminate if there was an error (for example target table
does not exists)

Extra after variables, for the gpload details:
		NB_LINE_INSERTED	Integer
		NB_LINE_UPDATED		Integer
		NB_DATA_ERRORS		Integer
		STATUS				String
		RUNTIME				String

Option to GZip compress the datafile, this will increase CPU usage, but
decreases DISK usage. (50-90% smaller files)
Remove data file after successful execution. If there were *any* errors
the data file is not removed.

## tGreenplumGPLoad/tGreenplumGPLoad_main.javajet
Added byte[] type support. Byte[] is converted to escaped OCTET
characters, which is compatible with all PostgreSQL versions from 8 to
13.

Fixed when newline / escape characters appeared in the data.

Postgresql uses null terminated bytes so those have to be removed from
the data to avoid errors.

## tGreenplumGPLoad/tGreenplumGPLoad_end.javajet
BugFix don't overwrite the actual inserted record number with 0

- [ ] Bugfix
- [x] Feature

- [ ] Yes
- [x] No

**Other information**:


pyzhou-talend (approved):
LGTM

PR #5827, opened 2021-02-05
"fix(TDI-45503):Comma missing for tELTMap with multiple inputs, when a…"

ovladyka:
Reverts Talend/tdi-studio-se#5826


ovladyka:
Unsuccessful.

PR #5882, opened 2021-03-01
fix(TUP-25109): backport TUP-25109

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-25109

**What is the new behavior?**
retrieve colums according to sql.

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/properties/controllers/GuessSchemaController.java
+++ b/main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/properties/controllers/GuessSchemaController.java
@@ -86,6 +86,7 @@ import org.talend.core.ui.properties.tab.IDynamicProperty;
 import org.talend.core.ui.services.ISQLBuilderService;
 import org.talend.core.utils.CsvArray;
 import org.talend.core.utils.KeywordsValidator;
+import org.talend.core.utils.TalendQuoteUtils;
 import org.talend.designer.core.i18n.Messages;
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/properties/controllers/GuessSchemaController.java
```diff
@@ -214,6 +215,11 @@ public Command createButtonCommand(Button btn, IContextManager manager) {
         IElementParameter elementParameterFromField = elem.getElementParameterFromField(EParameterFieldType.MEMO_SQL);
 
         memoSQL = (String) elementParameterFromField.getValue();
+        if (memoSQL == null || TalendQuoteUtils.removeQuotes(memoSQL).trim().isEmpty()) {
```
bgunics-talend, line 218: If the query is empty but we have a tablename then the query to be executed should be select * from schema.table; However this might not let that happen.

kjwang-talend (approved):

main/plugins/org.talend.designer.core/src/main/java/org/talend/designer/core/ui/editor/properties/controllers/GuessSchemaController.java
bhe-talendbj: Just discuss with our PO, both guess query and guess schema are based on input sql, so we will keep current behavior for now. We may enhance this later.

PR #5943, opened 2021-03-11
feat(TDI-45746): fix date context variable

ozhelezniak-talend:
* fix the logic of passing date variable to sub jobs

https://jira.talendforge.org/browse/TDI-45746

**What is the new behavior?**
the date context variables converting via string not via second as it was before

- [x] Bugfix

- [x] No


undx (approved):
LGTM

PR #6010, opened 2021-03-25
feat(TDI-45246): do not put in configuration hidden parameters

undx:
https://jira.talendforge.org/browse/TDI-45246

**Related PR:**
- https://github.com/Talend/tdi-studio-se/pull/6011


ozhelezniak-talend (approved):
LGTM

PR #6083, opened 2021-04-13
fix(TDI-45859): tFileCopy with different fs

chmyga:
* Add force copy parameter

* Copy and delete file instead of moving

If the file is being moved between filesystems the process might fail

**What is the new behavior?**
In case 'Force copy instead of move' parameter is checked we will copy the file and delete it

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:
https://jira.talendforge.org/browse/TDI-45859

### DIFF
--- a/main/plugins/org.talend.designer.components.libs/libs_src/filecopy/src/main/java/org/talend/FileCopy.java
+++ b/main/plugins/org.talend.designer.components.libs/libs_src/filecopy/src/main/java/org/talend/FileCopy.java
@@ -52,4 +52,21 @@ public class FileCopy {
         }
     }
 
+    /**
+     * Force Copy and Delete files.
+     *
+     * @param srcFileName : file name for source file.
+     * @param desFileName : file name for destination file.
+     * @throws IOException : if IO pb.
+     */
+    public static void forceCopyAndDelete(String srcFileName, String desFileName) throws IOException {
+        final Path source = Paths.get(srcFileName);
+        final Path destination =  Paths.get(desFileName);
+        final long lastModifiedTime = new File(srcFileName).lastModified();
+
+        Files.copy(source, destination, StandardCopyOption.REPLACE_EXISTING);
+        Files.delete(source);
+        new File(desFileName).setLastModified(lastModifiedTime);
+    }
+
 }
--- a/main/plugins/org.talend.designer.components.libs/libs_src/filecopy/src/main/java/org/talend/FileCopy.java
+++ b/main/plugins/org.talend.designer.components.libs/libs_src/filecopy/src/main/java/org/talend/FileCopy.java
@@ -66,7 +66,7 @@ public class FileCopy {
 
         Files.copy(source, destination, StandardCopyOption.REPLACE_EXISTING);
         Files.delete(source);
-        new File(desFileName).setLastModified(lastModifiedTime);
+        destination.toFile().setLastModified(lastModifiedTime);
     }
 
 }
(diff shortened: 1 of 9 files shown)


main/plugins/org.talend.designer.components.libs/libs_src/filecopy/src/main/java/org/talend/FileCopy.java
```diff
@@ -52,4 +52,21 @@ public static void copyFile(String srcFileName, String desFileName, boolean delS
...
+     * @throws IOException : if IO pb.
+     */
+    public static void forceCopyAndDelete(String srcFileName, String desFileName) throws IOException {
+        final Path source = Paths.get(srcFileName);
+        final Path destination =  Paths.get(desFileName);
+        final long lastModifiedTime = new File(srcFileName).lastModified();
+
+        Files.copy(source, destination, StandardCopyOption.REPLACE_EXISTING);
+        Files.delete(source);
+        new File(desFileName).setLastModified(lastModifiedTime);
```
clesaec, line 69: Use "destination.toFile().setLas..." instead of create new File instance.
But, on client example where copy was possible but not move, i thought it was change destination file attribute that was not allowed, like change modified time ?

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/SetForceCopyFortFileCopy.java
```diff
@@ -0,0 +1,67 @@
...
+        IComponentConversion action = new IComponentConversion() {
+
+            public void transform(NodeType node) {
+                if(node == null) {
+                    return;
+                }
+
+                if (ComponentUtilities.getNodeProperty(node, "FORCE_COPY_DELETE") == null) {//$NON-NLS-1$
+                    ComponentUtilities.addNodeProperty(node, "FORCE_COPY_DELETE", "CHECK");//$NON-NLS-1$ //$NON-NLS-2$
+                    ComponentUtilities.getNodeProperty(node, "FORCE_COPY_DELETE").setValue("true");//$NON-NLS-1$ //$NON-NLS-2$
```
clesaec, line 40: default value of FORCE_COPY_DELETE is false; why migration set it to true

clesaec (commented):
2 remarks

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/SetForceCopyFortFileCopy.java
chmyga: For all new jobs we need to leave it in default value. But for old jobs(where the default behavior was copy and delete) we need to set it to true, to make the behavior the same as before the changes.

main/plugins/org.talend.designer.components.libs/libs_src/filecopy/src/main/java/org/talend/FileCopy.java
chmyga: It looks to me that change modified time is not allowed for VM provided by support. But at the same time support says that customer has the modified time changed successfully. Anyway, if we can't move file it doesn't mean that we can't change the modified time for sure. Thus we should at least try to do that. The Files.setLastModifiedTime() method will throw an exception if it is not allowed, while this method will just return false. Also, this way of attribute modification was used before the global changes in our lib, thus it would work in the same way as before.
And thank you for the comment about "destination.toFile()". I'll update the PR.

clesaec (approved):
:+1:

clesaec (approved):
:+1:

PR #6095, opened 2021-04-15
fix(TDI-45577): Job using snowflake dynamic schema fails with special char

wwang-talend:
https://jira.talendforge.org/browse/TDI-45577

**What is the new behavior?**
https://jira.talendforge.org/browse/TDI-45577

- [x] Bugfix
- [ ] Feature

- [ ] Yes
- [x] No


jzhao-talend (approved):
LGTM

main/plugins/org.talend.designer.components.libs/libs_src/talend-codegen-utils/src/main/java/org/talend/codegen/enforcer/IncomingSchemaEnforcer.java
```diff
@@ -144,6 +149,12 @@ public void initDynamicColumn(String name, String dbName, String type, String db
         addDynamicField(name, type, null, format, description, isNullable);
     }
 
+    private Set<String> existNames;
```
wwang-talend, line 152: this works for avoiding duplicated name, but the scope only for dynamic fields, for example, if talend schema contains :
```
id
name
address
dynamic
```
only can't avoid duplicted in "dynamic" scope, can't avoid not duplicated with the basic column, but not important, we can fix it later as that is almost 1% case, no need to make more code for that.

PR #6290, opened 2021-06-01
feat(TBD-10185):Support Datasets API in tMap - Spark Batch - Inner Join / Match model: Unique match

lbourgeois:
Also activating Dataset when there is only one input connection.
It will cover the simple mapping case (no join in tMap)

### DIFF
--- a/main/plugins/org.talend.designer.mapper/src/main/java/org/talend/designer/mapper/MapperComponent.java
+++ b/main/plugins/org.talend.designer.mapper/src/main/java/org/talend/designer/mapper/MapperComponent.java
@@ -830,7 +830,7 @@ public class MapperComponent extends AbstractMapComponent implements IHashableIn
                 }
             }
             if (GlobalServiceRegister.getDefault().isServiceRegistered(IDesignerMapperService.class)) {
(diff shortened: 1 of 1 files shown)


lbourgeois:
will change branch name to separate unrelataed changes in Studio build

main/plugins/org.talend.designer.mapper/src/main/java/org/talend/designer/mapper/MapperComponent.java
```diff
@@ -880,18 +879,20 @@ public boolean isDatasetCompatible(boolean isJobValidForDataset) {
             return false;
         }
 
-        //only two input connections
-        if (this.externalData.getInputTables().size() != 2) {
+        // max two input connections
+        if (this.externalData.getInputTables().size() > 2) {
```
AlixMetivier, line 883: dataset is working with only one column in input ? I thought the !=2 was mandatory

lbourgeois: I thought I checked that simple mapping with no join was working in ds. Just rechecked and it is not. I will revert that part on https://github.com/Talend/tdi-studio-se/pull/6313/ (new PR with correct branch name)

PR #6960, opened 2021-12-01
feat(APPINT-32531) Make all DS/Route builds Java 11 and above compatible, design/build & run should work with different Java versions

apoltavtsev:


dsergent (approved):
LGTM

PR #6999, opened 2021-12-09
fix(TDI-47213): correct migration issue w/ context + function call

undx:
https://jira.talendforge.org/browse/TDI-47213
https://jira.talendforge.org/browse/TCOMP-2053

* Related PR*
- https://github.com/Talend/tdi-studio-se/pull/6999
- https://github.com/Talend/tdi-studio-se/pull/7010
- https://github.com/Talend/tdi-studio-se/pull/7009


clesaec (approved):
LGTM, 
but think to refactor later with WebSocketClient as an interface with sendAndWait method to allow very simple unit test on inner classes.

PR #7167, opened 2022-01-27
fix(TDI-47435): Compile errors when tPostgreSqlOutput enable 'Debug q…

dochkas-talend:
…uery mode' and 'action on data' is 'Upsert'

Modified some DB templates so they initialize unresolved variable when dataAction is 'Upsert'


dmytro-sylaiev (approved):
LGTM

PR #7303, opened 2022-03-11
fix(TCOMP-2003): Maven dependency classifier considered as version in dependencies.txt by Studio

undx:
https://jira.talendforge.org/browse/TCOMP-2003

**Related PRs**
- https://github.com/Talend/tdi-studio-se/pull/7302
- https://github.com/Talend/tdi-studio-se/pull/7303

* fix(TCOMP-2003): fix Mvn.toLocation in GAV parsing
(cherry picked from commit 8f5d517779880299a122c48e011f5de39e1f64a9)


clesaec (approved):
LGTM, even if I would prefer a MvnBuilder with methods like withGroup, withArtifact and a toMvn, but ok, the difficulties come from format it self, where values have no names.

PR #7305, opened 2022-03-14
fix(TDI-47560): tSybaseOutput BigDecimal jdbc inconsistencies

dochkas-talend:
Set BigDecimal for each Sybase version


main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
```diff
@@ -3351,7 +3351,11 @@ public class CLASS {
             } else {
                 return " null ";
             }
-        }        
+        }
+		private String dbVersion = "";
+        public void setDbVersion(String dbVersion){
```
ozhelezniak-talend, line 3356: * formatting issue
* imo you can place a new field and its method into the parent class it allows you to get rid of explicit cast.

main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
```diff
@@ -3478,9 +3482,13 @@ public class CLASS {
                 setStmt.append(prefix + cid + ".setNull(" + index + dynamic + ", java.sql.Types.TIMESTAMP);\r\n");
                 setStmt.append("}\r\n");
             } else if(typeToGenerate.equals("BigDecimal")){
-                if(column.getColumn().getPrecision()!=null && column.getColumn().getLength()!=null){
+                if(!"SYBSEIQ_16_SA".equals(dbVersion) && column.getColumn().getPrecision()!=null && column.getColumn().getLength()!=null){
```
ozhelezniak-talend, line 3485: What about SYBSEIQ_16_SA?
Didn't get the explanation on the ticket.

main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
dochkas-talend: I followed the example of OracleManager. There is exatly the same field and exactly the same cast there. If we want to move this field to the parent class then we have to move the field from OracleManager as well (it looks like a feature request)

main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
dochkas-talend: https://jira.talendforge.org/browse/TDI-7505
You can take a look a this ticket. It says that the cast is needed to throw correct exceptions. I checked it on SYBSEIQ_16_SA and all exceptions are thrown correctly because it uses another driver

ozhelezniak-talend: Cool. Just to be sure. In non exception\error case with BigDecimal works well?

main/plugins/org.talend.designer.components.localprovider/components/templates/db_output_bulk.skeleton
ozhelezniak-talend: Why does it look like feature request?

ozhelezniak-talend: I think we can easily move it.  With a comment that explains the purpose of this field. For ancestors.

ozhelezniak-talend (dismissed):
LGTM
Please ask also validation of Oracle DB (just in case, because officially we affected this part too)

ozhelezniak-talend (approved):
LGTM

PR #7324, opened 2022-03-18
fix(TDI-47539): tMicrosoftCRM change application secret as a password

dochkas-talend:
Changed field type and refactored code so that 2 passwords didn't conflict


dmytro-sylaiev (dismissed):

dmytro-sylaiev (approved):
LGTM

PR #7386, opened 2022-04-01
chore(TDI-47688): bump jackson

ozhelezniak-talend:

**What is the new behavior?**
bump jackson because of cve
it's a bit redundant because we manage the version of jackson in the project where it in use but to avoid confusion let's update it here too

* also removed un-used junit


undx (approved):
LGTM

PR #7410, opened 2022-04-08
fix(TDI-47168): Fix studio side metadata save

dmytro-sylaiev:

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/utils/SchemaUtils.java
+++ b/main/plugins/org.talend.designer.core.generic/src/main/java/org/talend/designer/core/generic/utils/SchemaUtils.java
@@ -46,6 +46,7 @@ import org.talend.cwm.helper.ConnectionHelper;
 import org.talend.cwm.helper.PackageHelper;
 import org.talend.cwm.helper.SchemaHelper;
 import org.talend.cwm.relational.RelationalFactory;
+import org.talend.daikon.avro.AvroNamesValidationHelper;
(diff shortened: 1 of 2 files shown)


clesaec (commented):
The [JIRA](https://jira.talendforge.org/browse/TDI-47168) is only about snowflake and code change is for all database. Is it general rules ?
As far as i understand [snowflake doc](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#double-quoted-identifiers)
`String fakeName = '"' + name + '"';`
would be enough.

And [AvroNamesValidationHelper.getAvroCompatibleName](https://github.com/Talend/daikon/blob/master/daikon/src/main/java/org/talend/daikon/avro/AvroNamesValidationHelper.java#L45)("my.identifier") will return value "_my.identifier", which is still not avro compatible, nor snowflake compatible. 
For the first, i agree, it's a bug from daikon, 
`AvroNamesValidationHelper.isValidParameterName( AvroNamesValidationHelper.getAvroCompatibleName("a.b"))  `
will return false

dmytro-sylaiev:
We can't make a fix only for snowflake: the pb here is that the avro schema's name we're saving to the metadata is used as a tableName in studio (not in jobs) and when the table name and metadata name differs, we got a regression. The original fix I did was to just alter the name of the table for the Snowflake Metadata, but when we edit the connection - the original table became unchecked. 
About the bug in daikon - good point. I think I'll create a jira bug to fix out of the scope of the original issue.

PR #7489, opened 2022-05-03
fix(TDI-47879): Removed destination property from tExtractJSONFields

dochkas-talend:
Removed property and added migration task to remove this property from old jobs


main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/RemoveDestinationPropertyExtractJSONFieldsMigrationTask.java
```diff
@@ -0,0 +1,61 @@
...
+                if (destination != null) {
+                    nodeType.getElementParameter().remove(destination);
+                }
+            }
+        };
+
+        IComponentFilter componentFilter = new NameComponentFilter(componentNameToAffect);
+
+        try {
+            ModifyComponentsAction.searchAndModify(item, processType,
```
ozhelezniak-talend, line 47: You should use the return value to check if smth was modified.

main/plugins/org.talend.repository/plugin.xml
```diff
@@ -3732,7 +3732,15 @@
               id="org.talend.repository.model.migration.EncryptClientSecretMicrosoftCRMMigrationTask"
               name="EncryptClientSecretMicrosoftCRMMigrationTask"
               version="8.0.1">
-
+        </projecttask>
+		<projecttask
+              beforeLogon="true"
+              breaks="8.0.1"
```
ozhelezniak-talend, line 3738: Not sure, maybe we need to use 8.0.0 here. Need to ask in tup-tdi.

ozhelezniak-talend (changes requested):
migration task should be modified. 
If the Migration task doesn't change anything it should return nothing to do.

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/RemoveDestinationPropertyExtractJSONFieldsMigrationTask.java
dochkas-talend: Ok, then I should also update my migration task from TDI-47539

ozhelezniak-talend: in the scope of another work item

main/plugins/org.talend.repository/plugin.xml
ozhelezniak-talend: Could you align tags?

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/EncryptClientSecretMicrosoftCRMMigrationTask.java
```diff
@@ -52,19 +52,21 @@ public void transform(NodeType node) {
...
         };
 
+        boolean modified = false;
+
         for (String componentName : componentsNameToAffect) {
             IComponentFilter componentFilter = new NameComponentFilter(componentName);
 
             try {
-                ModifyComponentsAction.searchAndModify(item, processType,
+                modified = modified || ModifyComponentsAction.searchAndModify(item, processType,
```
ozhelezniak-talend, line 61: || is a bad choice here.

ozhelezniak-talend: Also, I asked for a separate WorkItem. Don't mess with different changes for different tickets.

ozhelezniak-talend (changes requested):

main/plugins/org.talend.repository/src/main/java/org/talend/repository/model/migration/RemoveDestinationPropertyExtractJSONFieldsMigrationTask.java
```diff
@@ -0,0 +1,61 @@
...
+                    nodeType.getElementParameter().remove(destination);
+                }
+            }
+        };
+
+        IComponentFilter componentFilter = new NameComponentFilter(componentNameToAffect);
+
+        try {
+            boolean modified = ModifyComponentsAction.searchAndModify(item, processType,
+                    componentFilter, Arrays.asList(removeDestinationProp));
```
ozhelezniak-talend, line 48: You could use here Collections.singletonList

ozhelezniak-talend (approved):
lgtm

PR #8069, opened 2022-10-05
fix(APPINT-34581) - Route and data services features versions behave differently in Talend Runtime

sbieliaievl:


apoltavtsev (approved):
LGTM

PR #8474, opened 2023-02-09
fix(TDI-49107): Add support for input_row / output_row for tJavaFlex

dochkas-talend:
- add search for all input_row/output_row occurrences and replacement for current row name for main block of tJavaFlex


dmytro-sylaiev (approved):
LGTM

dmytro-sylaiev (approved):

PR #8582, opened 2023-03-21
fix(TUP-38314)Guess schema fails with the warning

jiezhang-tlnd:
https://jira.talendforge.org/browse/TUP-38314

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:


cmeng-talend (changes requested):
looks like it will cause junit failure, since there are many junits use mock property too, can find them by searching the call hierarchy of `AbstractGuessSchemaProcess.getNewmockProperty()`;
I think it's better to call something like the following codes when creating the guess schema process:
```
        IElementParameter log4jElemParam = process.getElementParameter(EParameterName.LOG4J2_ACTIVATE.getName());
        if (log4jElemParam != null) {
            log4jElemParam.setValue(false);
        }
```
Please take care that there are several guess schema functions for different usage:

1. standard component: GuessSchemaProcess.java

1. tacokit component: TaCoKitGuessSchemaProcess.java

1. dynamic guess schema feature: DynamicGuessSchemaProcess.java

maybe need to disable log4j for all these types, but also means that we should test all of them; and please also check whether there is junit failure after fixed.

cmeng-talend (approved):

PR #9131, opened 2023-09-28
APPINT-35418: Refer to wrong route in blueprint.xml when two routes use same routelet and deploy by CI

apoltavtsev:


arashankar (approved):
looks good to me!

PR #9190, opened 2023-10-25
Mhryb/tdi 49793 release streams in t map

mhryb:

**What is the new behavior?**

- [ ] Bugfix
- [ ] Feature

- [ ] Yes
- [ ] No

**Other information**:


main/plugins/org.talend.designer.components.localprovider/src/main/java/org/talend/designer/components/lookup/persistent/PersistentSortedLookupManager.java
```diff
@@ -535,11 +546,12 @@ public B next() throws IOException {
     }
 
     public void endGet() throws IOException {
-        for (ILookupManagerUnit<B> orderedBeanLookup : lookupList) {
+        for (ILookupManagerUnit<B> orderedBeanLookup : lookupCompleteList) {
             orderedBeanLookup.close();
         }
         clear();
         lookupList = null;
+        lookupCompleteList = null;
```
dmytro-sylaiev, line 554: after endGet you can not use the instance or PersistentSortedLookupManager like before, the `lookupList` is inited with new array in the initGet() method but `lookupCompleteList` only in the field initializer. Was it expected?

I don't fully understand 
1) why do we need both array and ArrayList
2) when we copy (addAll) the whole array to the arrayList content, might we have duplicates then?

PR #9450, opened 2023-12-29
feat(TCOMP-2283): Handle Dynamic column in guess schema

undx:
https://jira.talendforge.org/browse/TCOMP-2283


ypiel-talend (approved):
LGTM

